PR #35, opened 2021-04-12
WIP: Add basic Google Test infrastructure

greenscientist:
Using submodule to test the gtest library, based on
https://stackoverflow.com/questions/35998856/how-can-i-use-google-test-with-my-project-that-builds-via-autotools


tahini:
Included in #37, so closing it

PR #37, opened 2021-04-14
Gtestbase

tahini:


tahini:
@greenscientist I started experimenting with your patch and added some tests. If you have a chance, could you please take a look?

README.md
```diff
@@ -44,6 +44,17 @@ Then:
 make
 ```
 
+## Test
+
+trRouting uses [Googletest](https://github.com/google/googletest) to unit test the application. To run the tests, you must first fetch the googletest submodule once into the repo:
```
greenscientist, line 49: Strange. I believed that this should be automatic when cloning the repo. Maybe once we land the basic stuff

configure.ac
```diff
@@ -1,4 +1,4 @@
-AC_INIT([trRouting], [1.0], [TODO@TODO.PRG])
+AC_INIT([trRouting], [1.0], [https://github.com/kaligrafy/trRouting])
```
greenscientist, line 1: technically, this should be in a seprate commit

tests/cache_fetch/cache_fetcher_test.cpp
```diff
@@ -0,0 +1,39 @@
+#include "gtest/gtest.h" // we will add the path to C preprocessor later
```
greenscientist, line 1: This looks decent as a first test

configure.ac
tahini: sure!

tests/Makefile.am
```diff
@@ -0,0 +1,17 @@
+check_LTLIBRARIES = libgtest.la
+libgtest_la_SOURCES = ../googletest/googletest/src/gtest-all.cc
+libgtest_la_CPPFLAGS = -I$(top_srcdir)/googletest/googletest/include -I$(top_srcdir)/googletest/googletest
+libgtest_la_LDFLAGS = -pthread
+
+check_PROGRAMS = gtest
+
+gtest_SOURCES = gtest.cpp cache_fetch/cache_fetcher_test.cpp
```
tahini: My interrogation here is if we have to include all files from all subdirs here? I tried [briefly] to have a Makefile in the cache_fetch directory, but then I had to copy this whole Makefile.am there, but that means the gtest will be compile in each of those subdir? Anyway, it works for now, but I hope to add a looot more tests soon-ish.

tests/cache_fetch/cache_fetcher_test.cpp
tahini: > This looks decent as a first test

Hello-trRouting test, no stub data required and it highlights potentiel issues with default empty values!

tests/cache_fetch/cache_fetcher_test.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "parameters.hpp"
+#include "cache_fetcher.hpp"
+#include "cache_fetcher.cpp"
```
tahini: I haven't seen includes of cpp files very often... I did this to make it work. But there's probably some makefile magic to do so the build will find the cpp file. I'll have to dig into that. Either now, or open an issue for later.

tests/Makefile.am
greenscientist: Good questions, I'll need to dig into this

tests/cache_fetch/cache_fetcher_test.cpp
greenscientist: You need to link with the compiled .cpp file otherwise

tahini: "otherwise", meaning I have to add the cache_fetch.cpp file to the gtest_SOURCES in the Makefile.am above? Is there a way to declare a variable in the Makefile.am in src and use it here to add all the sources in src/ for instance? For now, I can just list it though...

greenscientist: A better option would be to build these CPP into a lib in src/Makefile.am and link to that lib in the test Makefile. (Instead of building directly into the executable. Should term, I'm fine with adding a TODO in this file and do it in a later commit. (Specially if you'd prefer that I do that part of  the work)

tests/cache_fetch/agencies_cache_fetcher_test.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "parameters.hpp"
+#include "cache_fetcher.hpp"
+#include "agency.hpp"
+#include "capnp/agencyCollection.capnp.h"
+
+class AgencyCacheFetcherFixtureTests : public ::testing::Test {
+protected:
+    TrRouting::Parameters params;
+    TrRouting::CacheFetcher cacheFetcher = TrRouting::CacheFetcher();
+    std::string invalidCustomPath = "invalidCacheFiles";
```
greenscientist: those should probably be const and the variable name capitalized?

src/agencies_cache_fetcher.cpp
```diff
@@ -59,10 +69,13 @@ namespace TrRouting
...
       //std::cout << TStr << ":\n" << Toolbox::prettyPrintStructVector(ts) << std::endl;
       close(fd);
-    }
-    else
-    {
-      std::cerr << "missing " << tStr << " cache file!" << std::endl;
+      return 0;
+    } catch (const std::exception& e) {
+      std::cerr << "Error opening cache file " << tStr << std::endl;
+      return -EBADMSG;
```
greenscientist, line 86: What's the rationale for EBADMSG?? (This is not a common error code, so a bit of comment about it would be nice

greenscientist (commented):
Small nits, but otherwise looks good. 

(Yes, we need to improve how we link with the main code, but we could improve later)

src/agencies_cache_fetcher.cpp
tahini: If you get there, it usually means that the serialized data cannot be deserialized correctly (version mismatch, reading problem, I got one case I think of the buffer too small to store the data, but I'm not sure, etc). What would make more sense? EINVAL?

tests/cache_fetch/cache_fetcher_test.cpp
tahini: You probably saw after writing this message, I managed to have it working with a separate Makefile in cache_fetch, so I think it's ok

src/agencies_cache_fetcher.cpp
tahini: I wish capnp would return something more meaningful than throw an error. At least, we will!

greenscientist: EINVAL would be a more common one for this case. If we want to have a different error code, EBADMSG could be ok, but we really need to have a comment to explain it and have it documented. 
(Usually, the MSG code apply to communication stuff. We are pretty close. It's fine to redefine meaning of errors code as long as we are close, but this need to be documented in the code. )

src/agencies_cache_fetcher.cpp
```diff
@@ -35,9 +37,21 @@ namespace TrRouting
 
     std::cout << "Fetching " << tStr << " from cache..." << " " << customPath << std::endl;
     
-    if (CacheFetcher::capnpCacheFileExists(cacheFileName + ".capnpbin", params, customPath))
+    std::string cacheFilePath = CacheFetcher::getFilePath(cacheFileName, params, customPath) + ".capnpbin";
+    if (!CacheFetcher::capnpCacheFileExists(cacheFilePath))
```
tahini: Since I do the check of the return type of fd below, I might as well not do this one and use the enoent return value of fd instead.

src/agencies_cache_fetcher.cpp
```diff
@@ -35,9 +37,21 @@ namespace TrRouting
...
     {
-      int fd = open((CacheFetcher::getFilePath(cacheFileName, params, customPath) + ".capnpbin").c_str(), O_RDWR);
+      std::cerr << "missing " << tStr << " cache file!" << std::endl;
+      return -ENOENT;
+    }
+
+    int fd = open(cacheFilePath.c_str(), O_RDWR);
+    if (fd < 0) {
```
tahini: This project uses allman braces!

tahini:
@greenscientist This PR is ready to review, for data robustness. It should not throw exceptions at load time anymore

src/nodes_cache_fetcher.cpp
```diff
@@ -81,56 +100,89 @@ namespace TrRouting
...
+
+        // put same node transfer at the beginning:
+        t->reverseTransferableNodesIdx.push_back(i);
+        t->reverseTransferableTravelTimesSeconds.push_back(0);
+        t->reverseTransferableDistancesMeters.push_back(0);
+        close(fd);
+      }
+      catch (const kj::Exception& e)
+      {
+        // TODO Do something about missing or faulty cache files?
```
tahini: This is a faulty file, return the error. And move this TODO where the missing files are checked

greenscientist:
All those return int really makes you like Rust more :)

src/persons_cache_fetcher.cpp
```diff
@@ -55,9 +56,23 @@ namespace TrRouting
         std::string filePath {cacheFilePath + ".capnpbin" + (filesCount > 1 ? "." + std::to_string(i) : "")};
         std::string cacheFilePath = CacheFetcher::getFilePath(filePath, params, customPath);
 
-        if (CacheFetcher::capnpCacheFileExists(cacheFilePath))
+        int fd = open(cacheFilePath.c_str(), O_RDWR);
+        if (fd < 0)
         {
-          int fd = open(cacheFilePath.c_str(), O_RDWR);
+          int err = errno;
```
greenscientist, line 62: All this should be in a utility function if you keep copy/pasting it.

greenscientist (approved):
As mentionned, some of the error handling code would benefit from being made into function instead of being copy/pasted. Otherwise looks good.

tahini:
By accepting and merging #41, this PR is now obsolete since it was included in it. Closing it then

PR #38, opened 2021-04-20
Add customEndAtSeconds and uuid fields to Period

tahini:
In transition, those fields are being added to allow to customize when a
period should end and the uuid field is useful to save the period in the
database.

### DIFF
--- a/include/capnp/line.capnp
+++ b/include/capnp/line.capnp
@@ -29,6 +29,8 @@ struct Period {
   numberOfUnits        @7 :Int16;
   trips                @8 :List(Trip);
   isFrozen             @9 :Int8;
+  customEndAtSeconds  @10 :Int32;
+  uuid                @11 :Text;
 }
 
 struct Schedule {


tahini:
DO NOT MERGE YET I'm putting the patch here, but I'd like to add some unit tests around it and wait to see if the uuid field is absolutely necessary (since it goes from database to the cache, but it does not need to be the other way around...)

greenscientist:
@tahini Is this still relevant ?

tahini:
Still relevant I guess, but didn't have time to look into it. Is the period even necessary? Or could we just keep trips hours?

greenscientist:
@tahini, we should just do this. Having the capnp file out of sync does not make sense. Either we remove them from the transition side or we merge this in TrRouting.

tahini:
@greenscientist There's still the question of what to do with those fields. Is there any use case to have the capnp file have more data that what is needed in trRouting? Or do we consider those files are "just" a cache and can/should contain the minimal information required to run? (especially if we plan to store them along with calculations for later re-use, it goes towards minimal information)

greenscientist:
From the TrRouting perspective, this file should be in sync. 
The question to remove the field should be in the Transition side of things. So if we don't settle this idea shortly, I propose we sync the files and figure it out later

PR #44, opened 2021-05-03
Remove osrm lib requirement

kaligrafy:


.gitignore
```diff
@@ -8,6 +8,7 @@
 /trRoutingTripBasedConfig.yml
 /osrm
 /cache
+/src/.dirstamp
```
greenscientist, line 11: What is that?

src/osrm_fetcher.cpp
```diff
@@ -18,151 +18,74 @@ namespace TrRouting
...
+    //std::cout << "numberOfDurations: " << responseJson["durations"][0].dump(2) << std::endl;
+    int numberOfDistances = responseJson["distances"].size();
+    //std::cout << "numberOfDistances: " << responseJson["distances"][0].dump(2) << std::endl;
+    
+    if (responseJson["durations"] != nullptr && responseJson["distances"] != nullptr && responseJson["durations"][0] != nullptr && responseJson["distances"][0] != nullptr)
+    {
+      int numberOfDurations = responseJson["durations"][0].size();
       //std::cout << "numberOfDurations: " << responseJson["durations"][0].dump(2) << std::endl;
-      int numberOfDistances = responseJson["distances"].size();
+      int numberOfDistances = responseJson["distances"][0].size();
```
greenscientist, line 70: I'm curious about this change, I struggle to see how it's related to removing the OSRMlib

README.md
```diff
@@ -9,8 +9,7 @@ With random origin and destination (multiple accessible stops at origin and dest
 [Trib-Based Algorithm (TBA)][2] (not yet released)
 
 ## Dependencies
-[Open Source Routing Machine (OSRM)][3] (must be installed separately, see [install and usage instructions in OSRM Wiki][4])
-use -DBUILD_SHARED_LIBS=ON to install libraries needed to compile trRouting
+[Open Source Routing Machine (OSRM)][3] (an osrm server with a walking profile must be running for the transit region while making queries to the trRouting server)
```
greenscientist: Maybe we should add a bit more details on what it's need to run?  Could be a link to another document

Dockerfile
```diff
@@ -1,5 +1,3 @@
-# We need the osrm-backend lib to build trRouting so we rely on a fork of the official image
```
greenscientist, line 1: This will break the Dockfile in the transition project. It's ok, raising so people are aware. I've created an issue in the project to adapt it when this is pushed to production.

src/osrm_fetcher.cpp
kaligrafy: Wow, I don't remember having changed that in a while... This needs more testing.

Dockerfile
kaligrafy: ok

README.md
kaligrafy: We could copy this readme part from transition

.gitignore
kaligrafy: I don't know, I tought it was a file created by the compiler...

kaligrafy: Yes indeed, Automake generates an empty .dirstamp file

src/osrm_fetcher.cpp
kaligrafy: fixed in bdbb9c540879991cb2226f4a5be93e19f0d8f9ad

README.md
kaligrafy: added

kaligrafy:
Please review again. Thanks!

.gitignore
greenscientist: ok, never seen that before, but ok!

src/osrm_fetcher.cpp
greenscientist: You still have the extra [0] insome of the line there. Was it broken before or it's still untested code?

greenscientist (changes requested):

greenscientist:
Requested change, just to have an answer to the addition of an array indexing in the json response. (You removed some of the lines, but there's other.

greenscientist:
Hey! You have merged without a PR approuval

src/osrm_fetcher.cpp
kaligrafy: The correct way is with the index 0. It was not broken before, because the variables were initialized outside the loop first and never used inside the loop.

greenscientist: It's not that there were never used. They are used inside the loop, but they were redefined. (I missed that part and your comment have confused me a little bit ;-) )

kaligrafy: So it is correct? Can we close this?

greenscientist:
You already pushed it...

tahini:
This PR broke compilation with ubuntu... I don't understand why...

PR #45, opened 2021-05-07
Add a launch.json vscode configuration for debug

tahini:
This file allows to run/debug trRouting from vscode.

### DIFF
--- a/.vscode/launch.json.example
+++ b/.vscode/launch.json.example
@@ -1,0 +1,52 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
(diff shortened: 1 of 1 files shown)


greenscientist:
This seems fine. Would there be any conflict with people vscode own configuration ??

tahini:
There might indeed be conflicts if some people already have this file. They can probably easily be fixed. What I don't like is that even though it is in the .gitignore, changes to the file will be tracked though it will say

```
The following paths are ignored by one of your .gitignore files:
.vscode
Use -f if you really want to add them.
```

Should I rather rename it to launch.json.example?

tahini:
A year and a half later, launch.json.example is here...

greenscientist (changes requested):
Can you update the commit message to explain that it,s an example that people need to copy??

greenscientist (approved):

PR #50, opened 2021-06-18
fix wrong directory import for households, persons, od trips and places

kaligrafy:
This is needed to be compatile with where json2capnp saves the cache
files

### DIFF
--- a/src/households_cache_fetcher.cpp
+++ b/src/households_cache_fetcher.cpp
@@ -47,7 +47,7 @@ namespace TrRouting
     {
       boost::uuids::uuid dataSourceUuid = iter->first;
 
-      std::string cacheFilePath {"dataSources/" + boost::uuids::to_string(dataSourceUuid) + "/households/" + cacheFileName};
+      std::string cacheFilePath {"dataSources/" + boost::uuids::to_string(dataSourceUuid) + "/" + cacheFileName};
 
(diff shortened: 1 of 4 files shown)


src/households_cache_fetcher.cpp
```diff
@@ -47,7 +47,7 @@ namespace TrRouting
     {
       boost::uuids::uuid dataSourceUuid = iter->first;
 
-      std::string cacheFilePath {"dataSources/" + boost::uuids::to_string(dataSourceUuid) + "/households/" + cacheFileName};
+      std::string cacheFilePath {"dataSources/" + boost::uuids::to_string(dataSourceUuid) + "/" + cacheFileName};
```
tahini, line 50: Does it mean that in the root of the cache, you'll potentially have N files named households.capnpbin.n? Same for the others? It made sense to have them in their own directory, like the lines and nodes. Maybe it's json2capnp that should be changed?

tahini (commented):
Just asking a question here. NOT an approval.

src/households_cache_fetcher.cpp
kaligrafy: Well, that is a good question, but I think for now, we just create one file per class and we will refactor the cache system later.

tahini: Then why bother with a count file in this case if they all go in a single file? Also, careful, the time when we have enough data to be worth splitting may come sooner than the later time when you plan to refactor the cache system ;-)

PR #51, opened 2021-06-18
When using od trips routing, initial lost time at departure was not

kaligrafy:
available in the routing results. This commit adds the variable so we
can export it for each routed od trip

### DIFF
--- a/connection_scan_algorithm/src/forward_journey.cpp
+++ b/connection_scan_algorithm/src/forward_journey.cpp
@@ -430,6 +430,7 @@ namespace TrRouting
             result.arrivalTimeSeconds            = arrivalTime;
             result.departureTimeSeconds          = minimizedDepartureTime;
             result.initialDepartureTimeSeconds   = departureTimeSeconds;
+            result.initialLostTimeAtDepartureSeconds = minimizedDepartureTime - departureTimeSeconds;
             result.numberOfTransfers             = numberOfTransfers;
             result.inVehicleTravelTimeSeconds    = totalInVehicleTime;
             result.transferTravelTimeSeconds     = totalTransferWalkingTime;
@@ -466,6 +467,8 @@ namespace TrRouting
       result.travelTimeSeconds           = -1;
       result.arrivalTimeSeconds          = -1;
       result.departureTimeSeconds        = departureTimeSeconds;
+      result.initialDepartureTimeSeconds = -1;
+      result.initialLostTimeAtDepartureSeconds = -1;
       result.numberOfTransfers           = -1;
       result.inVehicleTravelTimeSeconds  = -1;
       result.transferTravelTimeSeconds   = -1;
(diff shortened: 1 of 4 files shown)


connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -567,6 +571,7 @@ namespace TrRouting
       result.transferWaitingTimeSeconds  = -1;
       result.firstWaitingTimeSeconds     = -1;
       result.nonTransitTravelTimeSeconds = -1;
+      result.nonTransitTravelTimeSeconds = -1;
```
greenscientist: This is a duplicate of the previous line

include/routing_result.hpp
```diff
@@ -13,6 +13,7 @@ namespace TrRouting
     int arrivalTimeSeconds;
```
greenscientist, line 13: Could be interesting to make it a class, so we can have a proper initialization for all field. 
And not have to copy paste and keep in sync in multiple place

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -466,6 +467,7 @@ namespace TrRouting
       result.travelTimeSeconds           = -1;
       result.arrivalTimeSeconds          = -1;
       result.departureTimeSeconds        = departureTimeSeconds;
```
greenscientist, line 469: You dont have     result.initialDepartureTimeSeconds = -1;  here
￼

greenscientist (changes requested):
Need a few fix up.

connection_scan_algorithm/src/reverse_journey.cpp
kaligrafy: fixed

include/routing_result.hpp
kaligrafy: I'll open an issue for this

connection_scan_algorithm/src/forward_journey.cpp
kaligrafy: fixed

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -430,6 +430,7 @@ namespace TrRouting
             result.arrivalTimeSeconds            = arrivalTime;
             result.departureTimeSeconds          = minimizedDepartureTime;
             result.initialDepartureTimeSeconds   = departureTimeSeconds;
+            result.initialLostTimeAtDepartureSeconds = minimizedDepartureTime - departureTimeSeconds;
```
greenscientist, line 433: Do you really need to add this in the struct since you can compute it from 2 others values already stored?

kaligrafy: I think yes, because the end-user can download the csv including these fields and I want to make sure they calculate it correctly

greenscientist (approved):
Ok, will follow up task #78

PR #59, opened 2021-07-02
Add -fuse-ld=gold to link flags

### LINKED ISSUE #57: trRouting does not compile on Linux
Since commit 10af9dfd8d8a9ee4d750b9594000a970f2394479, compilation fails with errors like

```
/usr/bin/ld: ../connection_scan_algorithm/src/libcsa.a(program_options.o): in function `boost::program_options::basic_command_line_parser<char>::basic_command_line_parser(int, char const* const*)':
/usr/include/boost/program_options/detail/parsers.hpp:32: undefined reference to `boost::program_options::detail::cmdline::cmdline(std::vector<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const&)'
```

greenscientist:
This flag was previously inherited from the OSRM libs. Seems to be related to how the boost libraries
are builds.
Works on debian

Fixes: #57 

Should be tested on other platform

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,6 +42,9 @@ LDFLAGS+=" $BOOST_LDFLAGS $BOOST_REGEX_LIB $BOOST_FILESYSTEM_LIB $BOOST_IOSTREAM
 # Add various flags
 LDFLAGS+=" $CURSES_LIBS"
 LDFLAGS+=" $CAPNP_LIBS"
+# Seems to be require by the boost library, at least on debian/ubuntu.
+# TODO: Probably need to be wrapped in a test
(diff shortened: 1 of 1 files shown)


kaligrafy:
on macos: 
invalid linker name in argument '-fuse-ld=gold'

tahini:
On Ubuntu, it works

kaligrafy:
Gold is an ELF-only linker. OS X (macos) does not use ELF.

greenscientist:
This gold flag is probably a red herring. WIll investigate furter

PR #60, opened 2021-07-02
Use LTLIBRARY instead of LIBRARY to build libcsa

### LINKED ISSUE #57: trRouting does not compile on Linux

greenscientist:
The static lib was creating relocation issues when linking with boost libraries.

Fixes: #57

Tested on debian only

### DIFF
--- a/connection_scan_algorithm/src/Makefile.am
+++ b/connection_scan_algorithm/src/Makefile.am
@@ -1,8 +1,8 @@
 AM_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/connection_scan_algorithm/include 
 
-noinst_LIBRARIES = libcsa.a
+noinst_LTLIBRARIES = libcsa.la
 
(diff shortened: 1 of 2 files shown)


kaligrafy:
Compiles successfully on macos

tahini (approved):
Works in Ubuntu

PR #62, opened 2021-07-13
Add a new nearestOsrmNodeNetworkDistance to access and egress steps

kaligrafy:
Useful to know if the origin and/or destination is far from the osrm network.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -93,7 +93,7 @@ namespace TrRouting
     std::string             alternativesRouting();
     std::string             odTripsRouting();
     
-    std::vector<int>        optimizeJourney(std::deque<std::tuple<int,int,int,int,int,short,int>> &journey);
+    std::vector<int>        optimizeJourney(std::deque<std::tuple<int,int,int,int,int,short,int,int>> &journey);
 
     /**
      * The update*FromCache methods get the data from the cache
@@ -202,7 +202,7 @@ namespace TrRouting
   private:
     
     enum connectionIndexes : short { NODE_DEP = 0, NODE_ARR = 1, TIME_DEP = 2, TIME_ARR = 3, TRIP = 4, CAN_BOARD = 5, CAN_UNBOARD = 6, SEQUENCE = 7, LINE = 8, BLOCK = 9, CAN_TRANSFER_SAME_LINE = 10, MIN_WAITING_TIME_SECONDS = 11 };
-    enum journeyStepIndexes: short { FINAL_ENTER_CONNECTION = 0, FINAL_EXIT_CONNECTION = 1, FINAL_TRANSFERRING_NODE = 2, FINAL_TRIP = 3, TRANSFER_TRAVEL_TIME = 4, IS_SAME_NODE_TRANSFER = 5, TRANSFER_DISTANCE = 6 };
+    enum journeyStepIndexes: short { FINAL_ENTER_CONNECTION = 0, FINAL_EXIT_CONNECTION = 1, FINAL_TRANSFERRING_NODE = 2, FINAL_TRIP = 3, TRANSFER_TRAVEL_TIME = 4, IS_SAME_NODE_TRANSFER = 5, TRANSFER_DISTANCE = 6, NEAREST_NODE_DISTANCE_METERS = 7 };
     
     int              departureTimeSeconds;
     int              initialDepartureTimeSeconds;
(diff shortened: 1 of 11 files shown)


connection_scan_algorithm/include/calculator.hpp
```diff
@@ -225,20 +225,22 @@ namespace TrRouting
...
     std::vector<int> tripsEnterConnection; // index of the entering connection for each trip index 
     std::vector<int> tripsEnterConnectionTransferTravelTime; // index of the entering connection for each trip index 
     std::vector<int> tripsExitConnection; // index of the exiting connection for each trip index 
     std::vector<int> tripsExitConnectionTransferTravelTime; // index of the exiting connection for each trip index 
     std::vector<int> tripsEnabled; // allow/disallow use of this trip during calculation
     std::vector<int> tripsUsable; // after forward calculation, keep a list of usable trips in time range for reverse calculation
-    std::vector<std::tuple<int,int,int>> accessFootpaths; // pair: accessNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters
-    std::vector<std::tuple<int,int,int>> egressFootpaths; // pair: egressNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters
+    std::vector<std::tuple<int,int,int,int>> accessFootpaths; // pair: accessNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters, nearest osrm node distance from point (only for acees and egress journey steps))
```
greenscientist: typo in the comments

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -70,9 +70,9 @@ namespace TrRouting
       int totalDistance            { 0}; int distance               {-1};
       int inVehicleDistance        {-1}; int totalInVehicleDistance { 0}; int totalWalkingDistance { 0};
       int totalTransferDistance    {-1}; int accessDistance         { 0}; int egressDistance       { 0};
-      int journeyStepTravelTime    {-1}; int accessWalkingTime      {-1};
-      int transferTime             {-1}; int egressWalkingTime      {-1};
-      int waitingTime              {-1}; int accessWaitingTime      {-1};
+      int journeyStepTravelTime    {-1}; int accessWalkingTime      {-1}; int nearestNetworkNodeDistanceMeters { -1};
```
greenscientist, line 73: You don't really need to put those outside the for loop if you are initializing them at every loop

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -96,9 +96,9 @@ namespace TrRouting
         totalDistance               =  0; distance               = -1; 
         inVehicleDistance           =  0; totalInVehicleDistance =  0; totalWalkingDistance =  0;
         totalTransferDistance       =  0; accessDistance         =  0; egressDistance       =  0;
-        journeyStepTravelTime       = -1; accessWalkingTime      = -1; 
-        transferTime                = -1; egressWalkingTime      = -1; 
-        waitingTime                 = -1; accessWaitingTime      = -1; 
+        journeyStepTravelTime       = -1; accessWalkingTime      = -1; nearestNetworkNodeDistanceMeters = -1;
+        transferTime                = -1; egressWalkingTime      = -1; nearestOriginNodeDistanceMeters = -1;
+        waitingTime                 = -1; accessWaitingTime      = -1; nearestDestinationNodeDistanceMeters = -1;
```
greenscientist, line 101: Please, don't  put multiple statement on the same line. 
Should also have accessWaitingTime multiple time

src/osrm_fetcher.cpp
```diff
@@ -68,6 +68,7 @@ namespace TrRouting
       int numberOfDurations = responseJson["durations"][0].size();
       //std::cout << "numberOfDurations: " << responseJson["durations"][0].dump(2) << std::endl;
       int numberOfDistances = responseJson["distances"][0].size();
+      int nearestNetworkNodeDistanceMeters = (int)ceil((float)responseJson["sources"][0]["distance"]); // the distance in meters between point and nearest osrm routable node
```
greenscientist, line 71: Have  you tested this ? responseJson is probably a string, you cannot just convert with (float)

greenscientist (changes requested):
Would be great to make those std::tuple<int,int,int,int,int,short,int,int>  into actual object so that we have an idea of what they are. 

Did a quick lock, there's a few things to improve, might find more things  later

kaligrafy:
About the tuple, when I wrote this code, it was faster to use tuples instead of objects... We should test the performance to see if it is necessary to use a tuple

tahini:
A good set of benchmarks will allow us to test all this stuff real soon! @YasmineMoumou !!

greenscientist:
> About the tuple, when I wrote this code, it was faster to use tuples instead of objects... We should test the performance to see if it is necessary to use a tuple

Sure, but they can also be at least behind a typedef or a #define. Each individual int can also be a typedef. 
The aim of the comment was to point how hard is it to understand the code at the moment with all these tuple. Coupled with a general lack of comments, this is really tricky

src/osrm_fetcher.cpp
kaligrafy: the distance attribute is always a float value

greenscientist: Je vais devoir aller lire la documentation de nlohmann::json
But in addition, we should not use C style casting in C++

kaligrafy:
I opened specific issues for each comment. Can I rebase this in the mean time?

src/osrm_fetcher.cpp
kaligrafy: added an issue for this

connection_scan_algorithm/include/calculator.hpp
kaligrafy: fixed

connection_scan_algorithm/src/reverse_journey.cpp
kaligrafy: Added an issue for this

src/osrm_fetcher.cpp
greenscientist: What about
int nearestNetworkNodeDistanceMeters = responseJson["sources"][0]["distance"].get<int>();

kaligrafy:
I would let you guys try to test and fix the C style casting issue as I am not good enough in C++ to do that efficiently

greenscientist:
@kaligrafy, can you create an issue with what you were trying to achieve here ? The changed changed a lot, so we cannot just try to apply this PR. I'll close the PR then and we can try to reimplement it.

PR #67, opened 2021-07-20
Test for capnp binary presence in configure script

### LINKED ISSUE #64: Add test in configure for presence of capnp binary
Make sure that we have the capnp executable to be able to generate the associated source code with the capnp definition.

greenscientist:
We need the binary, not just the library to be able to generate the capnp objects
header files

Fix #64

Second commit also use the found path in the Makefile

Tested by changing the capnp binary name to something non existant. The configure
step failed properly then.


tahini (approved):
Works here!

PR #77, opened 2021-08-02
Draft: Add benchmark test

### LINKED ISSUE #53: Setup test for benchmarking
* Add a test to calculate routes multiple times
* Redirect timestamps results to a file local to the project
* Print the average duration of the full CSA algorithm

YasmineMoumou:
This PR closes #53.

Changes:
* Add a benchmark test that reproduces TrRouting CSA main program and evaluates a request
*  Log benchmark results to a file

Work in progress:
* Read multiple requests in a CVS file for benchmarking


tahini:
I will try to work on this and complete integration in this sprint

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
```diff
@@ -0,0 +1,232 @@
...
+    double results[nbIter];
+    for (int i = 0; i < nbIter; i++)
+    {
+      auto start = chrono::high_resolution_clock::now();
+      calculator->alternativesRouting();
+      auto end = chrono::high_resolution_clock::now();
+      results[i] = chrono::duration_cast<chrono::nanoseconds>(end - start).count() * 1e-9;
+    }
+    benchmarkResultsFile << "Execution time results: " << std::endl;
+    long long resultSum = 0;
```
tahini, line 194: double

tahini (changes requested):
Some high level comments on what is left to do for complete benchmarks:

1- This needs benchmark data, in the form of a complete cache, with the right scenario ID. I generated my own cache data and updated the scenario ID, but we should provide a link to a file containing the proper data. Or in some later improvement, we could provide cache data and a file with the queries to benchmark for this specific archive.

2- Tests are run using `make check`, like unit tests. We should have another target `make benchmarks` to run these, but they can still use the gtest framework. Since there's only one query for now, it's ok, it's not too long, but before adding more, this should be done.

3- These benchmarks bootstrap the benchmarking framework, so there is only one tests. We should eventually have a csv file with all the benchmarks to run.

tahini:
Replaced with #87 (a squash of those commits + the long long -> double modification, and a commit to make the tests pass)

PR #79, opened 2021-08-02
Add bird distance accessibilitity implementation

### LINKED ISSUE #55: Add implementation for accessibility to get bird distance instead of just osrm backend
Instead of getting the accessible node path from osrm for all queries, it should be possible to also get this information as bird distance. That could be useful for installations where there is no osrm-backend server, which is not trivial to install, and also for unit tests.

That will require:

* Adding a parameter to queries to tell trRouting to get the accessibility information using bird distances. That parameter would be false by default and in this case use the system's default engine.

* Add a global parameter to specify the default engine to use for accessiblity, so it can use bird distances by default if we know there will not be any osrm-backend.

YasmineMoumou:
This PR closes #55.

Changes:
* Add bird distance calculation option to OSMR fetcher

### DIFF
--- a/include/osrm_fetcher.hpp
+++ b/include/osrm_fetcher.hpp
@@ -24,18 +24,24 @@ namespace TrRouting
 {
   class OsrmFetcher
   {
-  
+
   public:
-    
     OsrmFetcher() {}
-    OsrmFetcher(std::string projectShortname) {
-      
+    OsrmFetcher(std::string projectShortname)
+    {
     }
-    
-    static std::vector<std::tuple<int,int,int>> getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters& params, bool reversed = false);
-    
+
+    static std::vector<std::tuple<int, int, int>> getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, bool reversed = false);
+
+  protected:
+    static std::vector<std::tuple<int, int, int>> getNodesFromBirdDistance(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, Parameters &params);
+    static std::vector<std::tuple<int, int, int>> getNodesFromOsrm(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, bool reversed);
+
+    static std::tuple<float, float> calculateLengthOfOneDegree(const Point &point);
+    static float calculateMaxDistanceSquared(Parameters &params);
+    static float calculateNodeDistanceSquared(const Point *node, const Point &point, const std::tuple<float, float> &lengthOfOneDegree);
   };
-  
+
 }
 
 #endif // TR_OSRM_FETCHER
--- a/src/osrm_fetcher.cpp
+++ b/src/osrm_fetcher.cpp
@@ -1,52 +1,81 @@
 #include "osrm_fetcher.hpp"
 #include "json.hpp"
 
-
 namespace TrRouting
 {
-  
-  std::vector<std::tuple<int,int,int>> OsrmFetcher::getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters& params, bool reversed)
+  std::vector<std::tuple<int, int, int>> OsrmFetcher::getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, bool reversed)
   {
-
-    std::vector<int>                     birdDistanceAccessibleNodeIndexes;
-    std::vector<std::tuple<int,int,int>> accessibleNodesFootpaths;
-
-    float lengthOfOneDegreeOfLongitude = 111412.84 * cos(point.latitude * M_PI / 180) -93.5 * cos (3 * point.latitude * M_PI / 180);
-    float lengthOfOneDegreeOflatitude  = 111132.92 - 559.82 * cos(2 * point.latitude * M_PI / 180) + 1.175 * cos(4 * point.latitude * M_PI / 180);
-    float maxDistanceMetersSquared     = (params.maxAccessWalkingTravelTimeSeconds * params.walkingSpeedMetersPerSecond) * (params.maxAccessWalkingTravelTimeSeconds * params.walkingSpeedMetersPerSecond);
-    float distanceMetersSquared;
-    float distanceXMeters;
-    float distanceYMeters;
-
-    
-   
-    if (params.debugDisplay)
-      std::cout << "osrm with host " << params.osrmWalkingHost << " and port " << params.osrmWalkingPort << std::endl;
-    
-    //std::cout << "mode = " << mode << " speed = " << defaultSpeedMetersPerSecond << " maxTravelTime = " << maxTravelTimeSeconds << " port = " << osrmPort << " host = " << osrmHost << " " << std::endl;
-    //std::string queryString = "GET /table/v1/" + mode + "/" + std::to_string(point.longitude) +  "," + std::to_string(point.latitude);
-    std::string queryString = "/table/v1/" + mode + "/" + std::to_string(point.longitude) +  "," + std::to_string(point.latitude);
-
-    
-
-    int i {0};
-    for (auto & node : nodes)
+    if (params.birdDistanceAccessibilityEnabled)
     {
-      //std::cerr << node.point.latitude << std::endl;
-      distanceXMeters       = (node->point.get()->longitude - point.longitude) * lengthOfOneDegreeOfLongitude;
-      distanceYMeters       = (node->point.get()->latitude  - point.latitude)  * lengthOfOneDegreeOflatitude ;
-      distanceMetersSquared = distanceXMeters * distanceXMeters + distanceYMeters * distanceYMeters;
+      return getNodesFromBirdDistance(point, nodes, params);
+    }
+    else
+    {
+      return getNodesFromOsrm(point, nodes, mode, params, reversed);
+    }
+  }
+
+  std::vector<std::tuple<int, int, int>> OsrmFetcher::getNodesFromBirdDistance(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, Parameters &params)
+  {
+    std::vector<std::tuple<int, int, int>> accessibleNodesFootpaths;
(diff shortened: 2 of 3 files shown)


src/osrm_fetcher.cpp
```diff
@@ -85,11 +116,33 @@ namespace TrRouting
...
     if (params.debugDisplay)
         std::cout << "fetched osrm footpaths (" << accessibleNodesFootpaths.size() << " foopaths found)" << std::endl;
     
     return accessibleNodesFootpaths;
   }
-   
+
+  std::tuple<float,float> OsrmFetcher::calculateLengthOfOneDegree(const Point& point)
+  {
+    float lengthOfOneDegreeOfLongitude = 111412.84 * cos(point.latitude * M_PI / 180) -93.5 * cos (3 * point.latitude * M_PI / 180);
```
tahini: I know you did not write these lines, but where do 111412.84 come from? and 93.5? and the numbers of the line below? If @kaligrafy can answer those questions, you may add it as comments in this method. Also the numbers could be extracted to constants with meaningful names

src/osrm_fetcher.cpp
```diff
@@ -4,46 +4,77 @@
...
-      std::cout << "osrm with host " << params.osrmWalkingHost << " and port " << params.osrmWalkingPort << std::endl;
+        std::cout << "fetched footpaths using bird distance (" << accessibleNodesFootpaths.size() << " foopaths found)" << std::endl;
     
-    //std::cout << "mode = " << mode << " speed = " << defaultSpeedMetersPerSecond << " maxTravelTime = " << maxTravelTimeSeconds << " port = " << osrmPort << " host = " << osrmHost << " " << std::endl;
-    //std::string queryString = "GET /table/v1/" + mode + "/" + std::to_string(point.longitude) +  "," + std::to_string(point.latitude);
-    std::string queryString = "/table/v1/" + mode + "/" + std::to_string(point.longitude) +  "," + std::to_string(point.latitude);
+    return accessibleNodesFootpaths;
+  }
+
+  std::vector<std::tuple<int,int,int>> OsrmFetcher::getNodesFromOsmr(const Point& point, const std::vector<std::unique_ptr<Node>>& nodes, std::string mode, Parameters& params, bool reversed)
```
tahini: rename to getNodesFromOsrm, not Osmr

include/osrm_fetcher.hpp
```diff
@@ -34,6 +34,14 @@ namespace TrRouting
     
     static std::vector<std::tuple<int,int,int>> getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters& params, bool reversed = false);
     
+    private:
+    static std::vector<std::tuple<int,int,int>> getNodesFromBirdDistance(const Point& point, const std::vector<std::unique_ptr<Node>>& nodes, Parameters& params);
+    static std::vector<std::tuple<int,int,int>> getNodesFromOsmr(const Point& point, const std::vector<std::unique_ptr<Node>>& nodes, std::string mode, Parameters& params, bool reversed);
```
tahini: osmr => osrm

src/osrm_fetcher.cpp
```diff
@@ -4,46 +4,77 @@
...
 namespace TrRouting
 {
-  
-  std::vector<std::tuple<int,int,int>> OsrmFetcher::getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters& params, bool reversed)
+  std::vector<std::tuple<int,int,int>> OsrmFetcher::getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>>& nodes, std::string mode, Parameters& params, bool reversed)
   {
+    if (params.birdDistanceAccessibilityEnabled) 
+    {
+      return getNodesFromBirdDistance(point, nodes, params);
+    } else
```
tahini: in this project, the else is on a separate line
```
}
else
{
```

src/osrm_fetcher.cpp
```diff
@@ -4,46 +4,77 @@
...
 
-    int i {0};
+    int i {0}; 
     for (auto & node : nodes)
     {
-      //std::cerr << node.point.latitude << std::endl;
-      distanceXMeters       = (node->point.get()->longitude - point.longitude) * lengthOfOneDegreeOfLongitude;
-      distanceYMeters       = (node->point.get()->latitude  - point.latitude)  * lengthOfOneDegreeOflatitude ;
-      distanceMetersSquared = distanceXMeters * distanceXMeters + distanceYMeters * distanceYMeters;
+      distanceMetersSquared = calculateNodeDistanceSquared(node->point.get(), point, lengthOfOneDegree);
```
tahini, line 66: A good part of this method is a copy of the getNodesFromBirdDistance. Could you call the getNodesFromBirdDistance, then loop through the the results to build the query string instead?

include/osrm_fetcher.hpp
```diff
@@ -34,6 +34,14 @@ namespace TrRouting
     
     static std::vector<std::tuple<int,int,int>> getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters& params, bool reversed = false);
     
+    private:
```
tahini: I haven't done much C++ recently, but do we need to add the private methods here in the .hpp? Or are they implementation details that need only be in the .cpp file? @greenscientist  may know better.

tahini (changes requested):
iiuc, in this PR, you are not really adding the bird distance implementation, but refactoring the code to allow the bird distance to be used to get reachable nodes. Please make sure to update the commit message to reflect that.

Also, you 2 commits are not really independent, you can merge them in one, using interactive rebase `git rebase -i HEAD~2` and fixup the last one into the previous one.

src/osrm_fetcher.cpp
kaligrafy: It is explained [here](https://physics.wooster.edu/Manz/sandcollection/swaplist/Latitude%20and%20Longitude.pdf), but these number do not represent a physicial value exactly, they are just constants in the equation.

kaligrafy: We don't add the last part 0.118 cos(5 theta)), since it would change the longitude less than +/-11.8 cm

kaligrafy: These are the factors of a Taylor series approximation

tahini: Are those numbers specific to our latitude? Or would this calculation work anywhere in the world?

kaligrafy: in theory it works anywhere

greenscientist: Yeah,  source for these numbers needs to be put at least in comment in the code. And idealy as constant.

include/osrm_fetcher.hpp
greenscientist: If they are part of the class, they will need to be in the .hpp file. 
But since they are static, they could be helper function only in the .cpp file also. 
Also it's really rare that we need to set something like this as private, protected is usually better, so they can be used by derived classes.

greenscientist: Also, purely math function should probably be put into an utility class outside of the osrm class

YasmineMoumou:
> iiuc, in this PR, you are not really adding the bird distance implementation, but refactoring the code to allow the bird distance to be used to get reachable nodes. Please make sure to update the commit message to reflect that.
> 
> Also, you 2 commits are not really independent, you can merge them in one, using interactive rebase `git rebase -i HEAD~2` and fixup the last one into the previous one.

I can just squash them at the end into one?

src/osrm_fetcher.cpp
YasmineMoumou: The method getNodesFromOsrm requires more than just what getNodesFromBirdDistance returns in the accessibleNodesFootpaths vector, it requires the node itself to form the query. To seperate the two i had to have the loop through the nodes in both methods. Initially there was only one loop, but i separated them following your comment about having an if statement that redirects to the correct implementation in the main getAccessibleNodesFootpathsFromPoint method. I can retract to having a singe loop but if i do that, I will not have what you initially asked for.

YasmineMoumou: To remove some the dupplication i can take out the lines: 

    auto lengthOfOneDegree = calculateLengthOfOneDegree(point);
    float maxDistanceMetersSquared = calculateMaxDistanceSquared(params);

and place them before the method calls in getAccessibleNodesFootpathsFromPoint, but then i have to add 2 more parameters to both methods and i think there are already a lot of parameters (5 params) in each.

src/osrm_fetcher.cpp
YasmineMoumou: I think a comment is more suiting here because otherwise there will be 10 constants all used only once in this method no?

include/osrm_fetcher.hpp
YasmineMoumou: These methods are only used in this class. I can make them protected to leave the possibility of them being used by derived classes open, though from what i know, i don't think it will be needed.

src/osrm_fetcher.cpp
tahini: Ok, that's already great improvement to be able to be osrm independent. Eventually, we may split the osrmFetcher into to accessibleNodes  implementations, one using osrm, the other not, but for now, it's enough.

src/osrm_fetcher.cpp
```diff
@@ -80,16 +109,39 @@ namespace TrRouting
...
-    
+      std::cout << "fetched osrm footpaths (" << accessibleNodesFootpaths.size() << " foopaths found)" << std::endl;
+
     return accessibleNodesFootpaths;
   }
-   
+
+  std::tuple<float, float> OsrmFetcher::calculateLengthOfOneDegree(const Point &point)
+  {
+    // First two terms of the Taylor series of the longitude and latitude length functions
```
tahini: It's worth adding a link to something here. I tried to google what you wrote and couldn't find anything satisfactory. This link may be more helpful? https://gis.stackexchange.com/questions/75528/understanding-terms-in-length-of-degree-formula

For something algorithmic and using constants like this, it is worth putting a link to documentation on where the terms come from. The people reading this code may come from all kinds of background and may not be familiar with the geography or the math. A simple explanation makes it all clear!

tahini (changes requested):
One last small comment to improve a comment ;-)

Otherwise, what have you tested this with?

include/osrm_fetcher.hpp
YasmineMoumou: Conclusion: We keep the math methods here because there are not used elsewhere

PR #80, opened 2021-08-26
Draft: Integrate open mp parallelization

YasmineMoumou:
This PR adds a pragma to the alternativeRouting function before the foor loop. This PR also assesses the duration of the algorithm with and without the pragma using the benchmark test. (Source: https://www.geeksforgeeks.org/openmp-hello-world-program/ and https://moodle.polymtl.ca/mod/folder/view.php?id=120285&fbclid=IwAR0y7hI2Sv7DfOnXAqDT5yRr2AnbXkax1gEkVVYimJuRf8ad7aPqVszXwcQ)

To test without or with parallelization simply comment or uncomment the pragma code in trRouting/connection_scan_algorithm/src/alternatives_routing.cpp:
image

To reproduce these results, make sure to add -fopenmp after every "g++" in the Makefiles...
* trRouting/connection_scan_algorithm/src/Makefile
* trRouting/src/Makefile
* trRouting/tests/connection_scan_algorithm/Makefile
* trRouting/tests/Makefile

Note: This is temporary, normally the -fopenmp will be added in the Makefile.am (and pushed on the repo) and the right Makefile will be generated without needing to be added by every user in their local Makefile.

It is necessary to add a symbolic link to the cache in the following folders : 
to run benchmarks: trRouting/tests/connection_scan_algorithm/cache
to debug: trRouting/cache

To debug, a launch.json file has to be added to specify the debug configuration. Finally, these tests have been made for the montreal map, for a different location, the longitude and latitude coordinates of origin and destination need to be changed in the transition app form and the trRouting/tests/connection_scan_algorithm/benchmark_CSA_test.cpp (line 130 and 133). 

Results by starting TrRouting in debug mode and using the Transition web app :
image

Duration without parallelization :
image

Duration with parallelization: 
To set the number of threads to X type in the terminal 'export OMP_NUM_THREADS=X'

2 threads:
image
3 threads:
image

4 threads:
image

Results by executing the benchmark tests with the command 'make check'
(found in the trRouting/tests/connection_scan_algorithm/benchmarkResults.txt file):

Duration without parallelization (with all the pragma code commented):
image

Duration with parallelization: 
To set the number of threads to X type in the terminal 'export OMP_NUM_THREADS=X'
-------- Problems of segmentation fault when using the benchmark test with parallelization ---------

Solution idea:
In the pragma, we need to specify which variables are private and which are shared --> it is encouraged to write
default(none) next to the pragma to force the developer to explicitly choose private or shared for each variable
more info: https://pages.tacc.utexas.edu/~eijkhout/pcse/html/omp-data.html
I believe choosing the right option for each variable will get rid of the race condition. To make the right choices, we need to find which variables cause the race condition by debugging using valgrind (https://www.valgrind.org/docs/manual/drd-manual.html#drd-manual.data-race-detection).

To setup valgrind:
sudo apt-get update -y
sudo apt-get install -y valgrind
To run valgrind when running benchmark test in debug mode:
https://github.com/microsoft/vscode-cpptools/issues/4531 (in progress)

2 threads:

3 threads:

4 threads:


kaligrafy:
Great work thanks!
Is it normal that the screenshot is not there?

YasmineMoumou:
> Great work thanks!
> Is it normal that the screenshot is not there?

Yes! I am adding them, i had a meeting with Geneviève to try to fix the problem with the CI (in the addBenchmarkTest and intergrateOpenMPParallelization branches)

YasmineMoumou:
> Great work thanks!
> Is it normal that the screenshot is not there?

Yes! I am adding them, i had a meeting with Geneviève to try to fix the problem with the CI (in the addBenchmarkTest and intergrateOpenMPParallelization branches)

greenscientist:
Things have changed a lot since then. We probably to not need this for me moment, we might need to investigate if there's some part of the algorithm that we could parallelize.

PR #87, opened 2021-12-03
Add benchmarks for the CSA algorithm

tahini:
This adds a test application to run benchmarks for the CSA algorithm. To
run it, simply run `make check` and it will build this application and
run the benchmarks.

The benchmarks require cache data to be available in the
`tests/connection_scan_algorithm/cache/demo_transition` directory. The
cache data to match the benchmark can be found here:
https://nuage.facil.services/s/ntXfzgfBEFDS7M2 Simply unzip in the
`tests/connection_scan_algorithm/cache` directory. It corresponds to the
STM's fall 2018 18S_S service.

For now, only one query is benchmarked and hard-coded in the file. The
results are outputted in the benchmarkResults.txt file.

TODO:

* Move this benchmark application outside the `make check` target to
instead have a `make bench` target.
* Parameterize cache queries to benchmark multiple queries for the given
cache. Could be a .csv file that comes with a cache archive.
* Document how to do benchmarks.

### DIFF
--- a/tests/connection_scan_algorithm/Makefile.am
+++ b/tests/connection_scan_algorithm/Makefile.am
@@ -1,0 +1,33 @@
+check_PROGRAMS = gtest
+
+gtest_SOURCES = gtest.cpp \
+    benchmark_CSA_test.cpp \
+    ../../src/cache_fetcher.cpp \
+    ../../src/agencies_cache_fetcher.cpp \
+    ../../src/nodes_cache_fetcher.cpp \
+    ../../src/services_cache_fetcher.cpp \
+    ../../src/scenarios_cache_fetcher.cpp \
+    ../../src/lines_cache_fetcher.cpp \
+    ../../src/data_sources_cache_fetcher.cpp \
+    ../../src/networks_cache_fetcher.cpp \
+    ../../src/stations_cache_fetcher.cpp \
+    ../../src/paths_cache_fetcher.cpp \
+    ../../src/calculation_time.cpp \
+    ../../src/trips_and_connections_cache_fetcher.cpp \
+    ../../src/households_cache_fetcher.cpp \
+    ../../src/persons_cache_fetcher.cpp \
+    ../../src/modes_initialization.cpp \
+    ../../src/od_trips_cache_fetcher.cpp \
+    ../../src/osrm_fetcher.cpp \
+    ../../src/places_cache_fetcher.cpp
+
+
+gtest_LDADD = $(top_srcdir)/tests/libgtest.la ../../connection_scan_algorithm/src/libcsa.la
+
+gtest_LDFLAGS = -no-pie -pthread -std=c++17 -lstdc++fs
+
+gtest_CPPFLAGS = -I$(top_srcdir)/googletest/googletest/include -I$(top_srcdir)/googletest/googletest -I$(top_srcdir)/include -I$(top_srcdir)/connection_scan_algorithm/include -pthread -std=c++17 -lstdc++fs
+
+TESTS = gtest
+
+
--- a/tests/connection_scan_algorithm/benchmark_CSA_test.cpp
+++ b/tests/connection_scan_algorithm/benchmark_CSA_test.cpp
@@ -1,0 +1,237 @@
+#include "gtest/gtest.h" // we will add the path to C preprocessor later
+#include "parameters.hpp"
+
+#include "server_http.hpp"
+#include "client_http.hpp"
+
+//Added for the json-example
+#define BOOST_SPIRIT_THREADSAFE
+
+#include <vector>
+#include <algorithm>
+#include <string>
+#include <iostream>
+#include <iterator>
+#include <curses.h>
+#include <bits/stdc++.h>
+#include <chrono>
+#include <iostream>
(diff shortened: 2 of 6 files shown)


tahini:
As the author of this PR, I can't approve, but I'm fine with @YasmineMoumou's contribution here. We can open issues for the work still left to do and work on that.

kaligrafy (approved):

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
```diff
@@ -209,8 +209,12 @@ TEST_F(BenchmarkCSATests, TestGetFilePath)
 
   Calculator calculator(algorithmParams);
 
-  if (!updateCalculatorFromCache(&calculator))
-    ASSERT_EQ(true, false);
+  if (!updateCalculatorFromCache(&calculator)) {
+    // TODO That statement is to make sure the test fail if the cache is not set, but it always fails on github, just let it pass for now.
```
greenscientist, line 218: Maybe there's an ENV variable that is set on githut that we could check to make it skip this, but we could always add it later. 
Can this cause an issue when running locally?

tests/connection_scan_algorithm/Makefile.am
```diff
@@ -0,0 +1,31 @@
...
+    ../../src/scenarios_cache_fetcher.cpp \
+    ../../src/lines_cache_fetcher.cpp \
+    ../../src/data_sources_cache_fetcher.cpp \
+    ../../src/networks_cache_fetcher.cpp \
+    ../../src/stations_cache_fetcher.cpp \
+    ../../src/paths_cache_fetcher.cpp \
+    ../../src/calculation_time.cpp \
+    ../../src/trips_and_connections_cache_fetcher.cpp \
+    ../../src/households_cache_fetcher.cpp \
+    ../../src/persons_cache_fetcher.cpp ../../src/modes_initialization.cpp \
```
greenscientist: Splits those lines too?

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
```diff
@@ -0,0 +1,232 @@
...
+#include <string>
+#include <iostream>
+#include <iterator>
+#include <curses.h>
+#include <bits/stdc++.h>
+#include <chrono>
+#include <iostream>
+#include <fstream> 
+
+#include <boost/optional.hpp>.evaluateRequests();
```
greenscientist: Maybe add a comment for the .evaluateRequests()  I imagine it's fine, but it's quite unusual for C++ code

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
```diff
@@ -0,0 +1,232 @@
...
+#include "cache_fetcher.hpp"
+#include "calculation_time.hpp"
+#include "parameters.hpp"
+#include "scenario.hpp"
+#include "calculator.hpp"
+#include "program_options.hpp"
+#include "benchmark_CSA_test.hpp"
+
+using namespace TrRouting;
+using namespace std;
```
greenscientist: using namespace std is always really evil. I would normally decline the review just for that. Since the code have been around for a while, I won't do it, but would be nice if it can be fixed.

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
```diff
@@ -0,0 +1,232 @@
...
+      return -1;
+    }
+    else if (calculator->countNodes() == 0)
+    {
+      std::cout << "no nodes";
+      return -1;
+    }
+    else if (calculator->countLines() == 0)
+    {
+      return -1;
```
greenscientist, line 103: Should those other error also have cout?

greenscientist (commented):
See comments

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
tahini: As mentioned in the TODO in the PR description, the benchmarks shouldn't be part of `make check`, but we should have a separate target to run them, so this is just temporary. Locally, it will simply return and do nothing, so not really an issue...

tests/connection_scan_algorithm/benchmark_CSA_test.cpp
tahini: All those boost includes don't appear used at all, I'll just remove them all

greenscientist:
Looks ok. 
Only thing I would add is that maybe we should put the comment about the location of the data in a README or comment and not just in a commit message. 

(The "The benchmarks require cache data to be available in the
`tests/connection_scan_algorithm/cache/demo_transition` directory. The
cache data to match the benchmark can be found here:
https://nuage.facil.services/s/ntXfzgfBEFDS7M2 Simply unzip in the
`tests/connection_scan_algorithm/cache` directory. It corresponds to the
STM's fall 2018 18S_S service."

greenscientist:
Should be good as a first step

tahini:
> Should be good as a first step

Is that an official approval to rebase and merge?

greenscientist:
Yes

PR #88, opened 2021-12-06
WIP: Add an OpenAPI documentation of the trRouting API

tahini:
This commit starts a `docs/API.yml` API documentation file using the
OpenAPI 3.0.2 specification. It describe the routing parameters, for now
only those used by the transition query in typescript.

Other parameters are listed in comments in the file, it is to be
discussed whether they are all required or not.

Responses for error and no_routing_found are documented, but the details
of the success responses are not there yet.

### DIFF
--- a/docs/API.yml
+++ b/docs/API.yml
@@ -1,0 +1,568 @@
+openapi: '3.0.2'
+info:
+  title: TrRouting
+  version: '1.0'
+servers:
+  - url: http://localhost:4000
+
+paths:
+  /route/v1/transit:
+    get:
+      description: Calculate transit route results from a given origin/destination, or ODTrip, or accessibility map
+      parameters:
+        - in: query
+          name: origin
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: false
+          description: Comma-separated latitude/longitude coordinates of the origin, in the WSG84 coordinates system
+        - in: query
+          name: destination
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: false
+          description: Comma-separated latitude/longitude coordinates of the destination, in the WSG84 coordinates system
+        - in: query
+          name: od_trip_uuid
+          schema:
+            type: string
+          required: false
+          description: UUID of the OD trip to query. Either this or origin and destination should be specified.
+        - in: query
+          name: scenario_uuid
+          schema:
+            type: string
+          required: true
+          # TODO: Scenario is very transition centric. Can't trRouting be used with a GTFS or some other type, without a scenario
+          description: UUID of the scenario to query. A scenario defines the services, agencies and lines to use for this transit calculation
+        - in: query
+          name: alternatives
+          schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+            default: 0
+          required: false
+          description: Whether the results should return various alternatives if available or just a single result.
+        - in: query
+          name: departure_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: departure time in seconds since midnight
+        - in: query
+          name: arrival_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: arrival time in seconds since midnight
+        - in: query
+          name: min_waiting_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: "The minimum time to wait at a stop/station, to cope with uncertainties in the vehicle arrival times. Suggested value: 180"
+        - in: query
+          name: max_access_travel_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, to reach the first stop/station in the trip"
+        - in: query
+          name: max_egress_travel_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, from the last stop/station, to reach the destination"
+        - in: query
+          name: max_transfer_travel_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, for each transfer between stop/station during the trip"
+        - in: query
+          name: max_travel_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: The maximum total travel time between origin and destination, including access, transfer and egress times
+        - in: query
+          name: max_first_waiting_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: The maximum time, in seconds, one can wait at first stop/station to consider this trip valid
+        - in: query
+          name: all_nodes
+          schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+            default: 0
+          required: false
+          # TODO: This will become 2 different endpoints since the return values are so much different.
+          description: If set to 0, the query returns the routes between origin and destination. If set to 1, the query will return all the nodes accessible with the given parameter. Only the origin is required for this query.
+      # The parameters above are those used in the Transition queries, already ported to typescript
+      # Still undocumented parameters. See if they are all required and remove any reference to them if they are not. Parameters on the same line are in the same if condition in the code, so identical
+      # Some parameters differ only with the unit, consider using a single parameter, and if really required, add a parameter that contains the units. Otherwise, one can specify 2 such parameters, which one to take?
+      # - starting_node_uuid, start_node_uuid, origin_node_uuid
+      # - ending_node_uuid, end_node_uuid,destination_node_uuid
+      # - return_all_nodes_results, return_all_nodes_result (same as all_nodes)
+      # - time, departure, departure_time, start_time (redundant with departure_time_seconds, different format)
+      # - arrival_time, arrival, end_time (redundant with arrival_time_seconds, different format)
+      # - departure_seconds, start_time_seconds (same as departure_time_seconds)
+      # - arrival_seconds, end_time_seconds (same as arrival_time_seconds)
+      # - data_source_uuid
+      # - reverse
+      # - min_waiting_time, min_waiting_time_minutes (redundant with min_waiting_time_seconds, different units)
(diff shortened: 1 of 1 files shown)


docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+    get:
+      description: Calculate transit route results from a given origin/destination
+      parameters:
+        - in: query
+          name: origin
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: false
+          description: Comma-separated latitude/longitude coordinates of the origin
```
greenscientist: Should probably specify in which coordinate system. I imagine it will be the one from the underlying OSRM ?)

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+        - in: query
+          name: destination
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: false
+          description: Comma-separated latitude/longitude coordinates of the destination
```
greenscientist: Same

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+          schema:
+            type: string
+          required: false
+          description: UUID of the OD trip to query. Either this or origin and destination should be specified.
+        - in: query
+          name: scenario_uuid
+          schema:
+            type: string
+          required: true
+          description: UUID of the scenario to query
```
greenscientist: What's a scenario?

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+        - in: query
+          name: alternatives
+          schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+          required: false
+          description: Whether the results should return various alternatives if available or just a single result.
```
greenscientist, line 49: Should we document the default value here? I'm not sure what's the expectation of OpenAPI for that

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+        - in: query
+          name: departure_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: departure time in seconds since midnight
```
greenscientist, line 55: Just a generic question, but how do we define the timezone?

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, from the last stop/station, to reach the destination"
+        - in: query
+          name: max_transfer_travel_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, for the transfer between stop/station during the trip"
```
greenscientist: Is the maximum for each transfers or for all the transfer. Would be good to specify

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+        - in: query
+          name: max_travel_time_seconds
+          schema:
+            type: integer
+          required: false
+          description: The maximum total travel time between origin and destination, including access and egress times
```
greenscientist: Is it worth to add the transfer time in the description or it's implicit enough ?
"including access, egress and transfer time" ?

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+          description: The maximum time, in seconds, one can wait at first stop/station to consider this trip valid
+        - in: query
+          name: all_nodes
+          schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+          required: false
+          description: If set to 1, the query will return all the nodes accessible with the given parameter. Only the origin is required for this query.
```
greenscientist: What is the data returned if set to false

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+      # The parameters above are those used in the Transition queries, already ported to typescript
+      # Still undocumented parameters. See if they are all required and remove any reference to them if they are not. Parameters on the same line are in the same if condition in the code, so identical
+      # Some parameters differ only with the unit, consider using a single parameter, and if really required, add a parameter that contains the units. Otherwise, one can specify 2 such parameters, which one to take?
```
greenscientist, line 111: Yeah, having same parameters with different units is weird,

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+          name: all_nodes
+          schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+          required: false
+          description: If set to 1, the query will return all the nodes accessible with the given parameter. Only the origin is required for this query.
+      # The parameters above are those used in the Transition queries, already ported to typescript
+      # Still undocumented parameters. See if they are all required and remove any reference to them if they are not. Parameters on the same line are in the same if condition in the code, so identical
```
greenscientist, line 110: You are not kinding when you say there's a lot of them!

docs/API.yml
```diff
@@ -0,0 +1,228 @@
...
+      # - num_batches
+      # - debug
+      responses:
+        '200':
+          description: OK
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/error'
```
greenscientist, line 175: Error should be a 4xx return code probably (that's what OSRM does)

greenscientist (commented):
Good first pass. 

Added a few comments about the doc. Some times it's comment about the API, I imagine we would fix them later, but might be good to leave a trace as TODO maybe

docs/API.yml
tahini: I assume WSG84 (lat/lon). But @kaligrafy, can it be something else? Could the cache be of different format, then it would need to match the cache data?

docs/API.yml
tahini: I don't think trRouting needs to manage it. TrRouting is usually very local for now, but it's true that if we model the world, we'll have to know which midnight to take... d'oh! For now, it assumes one timezone though.

docs/API.yml
tahini: I'll do that for the new endpoints

tahini (commented):
Thanks for the review, I'll update and add documentation for the first new endpoint.

docs/API.yml
greenscientist: Yeah, probably need need at the moment, we'll need to plan for it in the future. Timezone management always become an issue when you talk about time!

docs/API.yml
tahini: Actually, not entirely true, the 4xx errors are invalid errors from the client. Only part of the error schema could be in the 4xx schema, like missing scenarios, or invalid coordinates.  But errors like error reading the cache, or missing data, etc fit under 200 code, with proper error code

docs/API.yml
```diff
@@ -0,0 +1,690 @@
...
+              schema:
+                oneOf:
+                  # TODO Some of the error schemas should be under a 4xx response, but that is not how it is in this endpoint
+                  - $ref: '#/components/schemas/error'
+                  - $ref: '#/components/schemas/no_routing_found'
+                  - $ref: '#/components/schemas/success'
+                  - $ref: '#/components/schemas/data_error'
+                discriminator:
+                  propertyName: status
+  /v1/routeOD:
```
tahini: I don't like this route's name. Suggestions? v1/route/od, with others for v1/route/access (to replace allNodes = 1, for accessibility map) and v1/route/odtrip ?

tahini:
For this PR, I'm keeping only the API description of the current route, with complete return types of the origin/destination calculation (and alternatives) requests. The new endpoint will come along with its implementation, in a separate PR.

docs/API.yml
greenscientist: Maybe error reading the cache should be a 5xx error? (Not super important for the moment, but it's cool to reuse some standard

docs/API.yml
```diff
@@ -0,0 +1,568 @@
...
+            transferWaitingTimeSeconds:
+              type: number
+              description: same as transferWaitingTimeMinutes, in seconds
+            firstWaitingTimeMinutes:
+              type: number
+              description: Time spent waiting at the first transit stop of the trip, in minutes
+            firstWaitingTimeSeconds:
+              type: number
+              description: same as firstWaitingTimeMinutes, in seconds
+            totalWaitingTimeMinutes:
```
greenscientist, line 370: These returns as Minutes and Seconds are very annoying. The minute one should be deprecated! :)

greenscientist (approved):
It looks good from my perspective. Good enough for documentation at the moment
maybe @kaligrafy want to also have a look?

kaligrafy:
I would like to change the order of lat,lon to lon,lat to follow geojson and standard x,y order, but that is a breaking change and we must change the order everywhere... How should we do this?

greenscientist:
> I would like to change the order of lat,lon to lon,lat to follow geojson and standard x,y order, but that is a breaking change and we must change the order everywhere... How should we do this?

First, open an issue for it

docs/API.yml
kaligrafy: It  should also work if we provide a destination. It will give you the departure time at each stop

tahini:
> I would like to change the order of lat,lon to lon,lat to follow geojson and standard x,y order, but that is a breaking change and we must change the order everywhere... How should we do this?

In the next patches, I'm proposing a new endpoint for a single route query, with only the necessary parameters. I can change the order for this new endpoint and leave this one unchanged to not have to modify current code.

kaligrafy:
that sounds good, like using v2 in the route?

kaligrafy:
when routing with all nodes, a destination can also be provided, and it will calculate from all stops to this destination.

tahini:
I want to split the all_nodes and route queries (and od trip too): they don't require the same parameters and they don't have the same return value. For the API, we want it clear and make sure any API checker can validate there are no mistakes. So `route` requires an origin/destination, while the `all_nodes` requires only one geographical parameter and have a parameter to say if it's origin or destination.

How the code handles the requests is an implementation detail. It can stay as it is now without problem

tahini:
I updated the description of the all_nodes parameter to mention that either origin or destination can be specified.

kaligrafy (approved):

PR #89, opened 2021-12-06
Cache the response for invalid data

tahini:
When the data from the cache is invalid or insufficient, the response
will always be the same, it can thus be cached to avoid recalculating
every time and all endpoints can use it.

### DIFF
--- a/connection_scan_algorithm/src/transit_routing_http_server.cpp
+++ b/connection_scan_algorithm/src/transit_routing_http_server.cpp
@@ -40,6 +40,43 @@ std::string consoleCyan       = "";
 std::string consoleMagenta    = "";
 std::string consoleResetColor = "";
 
+std::string intializeResponse(int dataStatus, Calculator& calculator)
+{
+  std::string response {""};
+  if (dataStatus < 0) {
+    response = "{\"status\": \"data_error\"}";
+  }
+  else if (calculator.countAgencies() == 0)
+  {
+    response = "{\"status\": \"error\", \"error\": {\"error\": \"No agencies found\", \"code\": \"MISSING_DATA_AGENCIES\"}}";
+  }
+  else if (calculator.countServices() == 0)
+  {
+    response = "{\"status\": \"error\", \"error\": {\"error\": \"No services found\", \"code\": \"MISSING_DATA_SERVICES\"}}";
+  }
+  else if (calculator.countNodes() == 0)
+  {
+    response = "{\"status\": \"error\", \"error\": {\"error\": \"No nodes found\", \"code\": \"MISSING_DATA_NODES\"}}";
+  }
+  else if (calculator.countLines() == 0)
+  {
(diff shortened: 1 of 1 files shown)


connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -99,6 +99,41 @@ int main(int argc, char** argv) {
   std::cout << "preparing calculator..." << std::endl;
   int dataStatus = calculator.prepare();
 
+  // Prepare default response if data is invalid or not complete
+  char* invalidResponse = nullptr;
```
greenscientist: Seems a bad practice to use char* in C++

Also can all of this be in a function?

greenscientist (changes requested):

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -286,41 +326,12 @@ int main(int argc, char** argv) {
...
-      response = "{\"status\": \"error\", \"error\": {\"error\": \"No paths found\", \"code\": \"MISSING_DATA_PATHS\"}}";
-    }
-    else if (calculator.countScenarios() == 0)
-    {
-      response = "{\"status\": \"error\", \"error\": {\"error\": \"No scenarios found\", \"code\": \"MISSING_DATA_SCENARIOS\"}}";
-    }
-    else if (calculator.countConnections() == 0  || calculator.countTrips() == 0)
-    {
-      response = "{\"status\": \"error\", \"error\": {\"error\": \"No schedules found\", \"code\": \"MISSING_DATA_SCHEDULES\"}}";
+    if (!invalidResponse.empty()) {
```
tahini: I'm not sure what to do now. I just wanted to avoid copy pasting the code above in the new endpoint, but anyway, the responses will be a bit different, no error object just error codes, so I might as well forget about this PR and copy paste for the new endpoint.

Besides, now the server needs to be restarted when the cache is updated, but there is an untested feature to just reload, so the new cache data may become OK in a later update, so we would need to re-instanciate a new server object after cache update? That would be OK too...

I don't have any strong opinion here.

greenscientist: You can put at least the string generation in a function. Then you can either just reuse it or cache it if you think it would be useful.

greenscientist (commented):
I would suggest something drasticly different, feel free to use or reject. 

Calculator should have a fonction "testValidatity" or something which return an error code. 
Or a isValid and a GetError functions. 
This would return a error code. 
You could then have a function which convert the ErrorCode into a String. 
Then a function which build an error response string. 

Anyway, just some thoughts

tahini:
I didn't want to touch the calculator just yet, but I think I'll have to anyway for the parameter updates. This will be part of a bigger PR later (that I'll do my best to split for easier review :wink:  )

PR #94, opened 2021-12-09
Add a basic unit tests for the CSA algorithm

tahini:
This moves the benchmarks to its own directory, leaving the connection_scan_algorithm directory available for unit tests.

It adds basic unit test data, with one agency and 2 lines.

This PR bootstraps the test, prepares the data set and has only one successful calculation with basic parameters. More tests will be added soon, to test various use cases


kaligrafy (approved):

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -133,6 +133,9 @@ namespace TrRouting
     long long         countConnections();
     int               countNetworks();
 
+    // These functions allow to prepare internal calculation data once the node and schedule data are ready
+    void nodeDataImported();
```
greenscientist: I don't think the past tense is appropriate for an "action" function. Maybe importNodeData instead?

connection_scan_algorithm/src/initializations.cpp
```diff
@@ -9,6 +9,105 @@ namespace TrRouting
     algorithmCalculationTime = CalculationTime();
 
   }
+
+  void Calculator::nodeDataImported() {
```
greenscientist: I don't see any importation in this function. Is  the name appropriate?

connection_scan_algorithm/src/initializations.cpp
```diff
@@ -9,6 +9,105 @@ namespace TrRouting
...
+    std::cout << forwardConnections.size() << " connections" << std::endl; 
+
+    //int benchmarkingStart = algorithmCalculationTime.getEpoch();
+
+    int lastConnectionIndex = forwardConnections.size() - 1;
+
+    forwardConnectionsIndexPerDepartureTimeHour = std::vector<int>(32, -1);
+    reverseConnectionsIndexPerArrivalTimeHour   = std::vector<int>(32, lastConnectionIndex);
+    
+    int hour {0};
```
greenscientist: I know you only copy/pasted the code, but I wonder why those initialization are not in the for()

connection_scan_algorithm/src/initializations.cpp
```diff
@@ -9,6 +9,105 @@ namespace TrRouting
...
+    int i = 0;
+    for (auto & connection : forwardConnections)
```
greenscientist: Should  those for loops also splited on their only functions?

greenscientist: Optional but if you have a chance, you can comment out on what they do

connection_scan_algorithm/include/calculator.hpp
greenscientist: Should those function really be public? What happen if you call them at any time?

tests/connection_scan_algorithm/csa_simple_calculation_test.cpp
```diff
@@ -0,0 +1,42 @@
+#include <errno.h>
+#include <boost/uuid/uuid.hpp>
+#include <boost/uuid/uuid_io.hpp>
+
+#include "gtest/gtest.h" // we will add the path to C preprocessor later
```
greenscientist: I don't  understand this comment.

tests/connection_scan_algorithm/csa_test_base.hpp
```diff
@@ -0,0 +1,64 @@
...
+    inline static const boost::uuids::uuid pathSNUuid = uuidGenerator("2f31893e-b2fb-4ebe-bf1b-2ee77fe1f96b");
+    inline static const boost::uuids::uuid pathEWUuid = uuidGenerator("50f8e656-99b5-4ddd-b41f-b4804fe599f3");
+
+    inline static const boost::uuids::uuid trip1SNUuid = uuidGenerator("89f1ee54-1284-4861-939a-3f3241a3e461");
+    inline static const boost::uuids::uuid trip2SNUuid = uuidGenerator("71a1f3eb-e98c-49bc-b578-7d2e490e1f2f");
+    inline static const boost::uuids::uuid trip1EWUuid = uuidGenerator("af0a429f-84f3-4a83-9883-f6387d445cd8");
+    inline static const boost::uuids::uuid trip2EWUuid = uuidGenerator("ea8c90ba-7deb-46af-87c6-8af211e8fab7");
+
+    TrRouting::Parameters params;
+    TrRouting::Calculator calculator;
```
greenscientist, line 52: It feels strange the have those those object as part of the test object. I think it would be better if they are created as part of the setup(). It might be a side effect of the current design of the TrRouting code.

tests/connection_scan_algorithm/csa_test_base.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    // Setup functions. The base class will define a complete default data set,
+    // but specific test cases can override these methods and add their own data
```
greenscientist, line 55: Nice

greenscientist (changes requested):
Main issue is the name of the function in the second commit. 

Maybe the test data could live in it's own commit with a more detailled explaination on how it was generated, but I can be wrong on this.

connection_scan_algorithm/include/calculator.hpp
tahini: They should not, they're public only because of the unit tests! Most of the fields in this class can be split into different structures/classes: Data objects (agencies, lines, corresponding maps, etc), calculation objects (the ones that need to be modified by those 2 methods), parameter objects (I'm working on it in a separate branch)

If you call them anytime, it will just reset the nodes fields. You don't do that during a calculation, it's not thread safe, but we have a single thread, so...

tahini: These functions do not import, but are to be called after the node data is imported, whence the past tense... I could put them both in a single method and call it resetInternalData. It could also be part of the current reset method: if the size of the array modified and the nodes do not match, reset them all.

greenscientist: Yeah, their name should reflect what they do and not in what stated the object is

tests/connection_scan_algorithm/csa_simple_calculation_test.cpp
tahini: haha me neither, but it's everywhere we have this include because we always copy-paste it!!

greenscientist: Yeah, I was afraid that I wrote that down originally and not remember what it meant

tahini:
Thanks for the feedback

> Maybe the test data could live in it's own commit with a more detailled explaination on how it was generated, but I can be wrong on this.

By test data, you mean the gtfs files? Or the base_csa_test.cpp which adds the data object?

greenscientist:
> > Maybe the test data could live in it's own commit with a more detailled explaination on how it was generated, but I can be wrong on this.
> 
> By test data, you mean the gtfs files? Or the base_csa_test.cpp which adds the data object?

I was thinking of the content of tests/connection_scan_algorithm/unit_test_data_gtfs/

But the more I think about it, I'm not sure I would know how to go from the data to the data object in the code. Maybe the code need more comments.

tahini:
> I was thinking of the content of tests/connection_scan_algorithm/unit_test_data_gtfs/
> 
> But the more I think about it, I'm not sure I would know how to go from the data to the data object in the code. Maybe the code need more comments.

I put the data there because one can easily zip those files (I didn't want to put the zip binary files in git, so it's unzipped here) and import it in Transition to visualize the nodes and network and add some stuff. But it is not required. The unit test data should remain manual and manually modifiable.

greenscientist:
> > I was thinking of the content of tests/connection_scan_algorithm/unit_test_data_gtfs/
> > But the more I think about it, I'm not sure I would know how to go from the data to the data object in the code. Maybe the code need more comments.
> 
> I put the data there because one can easily zip those files (I didn't want to put the zip binary files in git, so it's unzipped here) and import it in Transition to visualize the nodes and network and add some stuff. But it is not required. The unit test data should remain manual and manually modifiable.

Yeah, it's good that way. I juste think we miss some documentation about the steps you have put to go from there to the data in the tests. (and maybe how to reimport in transition and do modification). 
I'm trying to think about our next intern who will to do more tests :)

tests/connection_scan_algorithm/csa_test_base.cpp
```diff
@@ -0,0 +1,341 @@
...
+ * - One inbound path per line
+ * - One service
+ * - 2 trips per path, one of which is cadenced to make transfers easier.
+ *
+ * For easy updates and manipulations for other test cases, the textual gtfs
+ * files are provided in the unit_test_data_gtfs directory. The original data
+ * was arbitrarily generated in Transition, then the time numbers were rounded
+ * to times more readable by a test case developer. The distances data was kept
+ * as original because some bird distance calculations actually do take the
+ * geographical data into account.
```
tahini: I will add some documentation here on what to do with this to re-import in Transition

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -191,6 +194,11 @@ namespace TrRouting
...
     std::vector<std::vector<std::unique_ptr<float>>> tripConnectionDemands;        // tripIndex: [connectionIndex (sequence in trip): sum of od trips weights using this connection (demand)]
     //std::vector<std::unique_ptr<std::vector<std::unique_ptr<int>>>>   tripIndexesByPathIndex;
+
+    // Made public for unit tests, they should not be, add helper methods instead
+    // TODO As part of issue https://github.com/chairemobilite/trRouting/issues/95, this will be removed
+    std::vector<std::shared_ptr<std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>>> forwardConnections; // tuple: departureNodeIndex, arrivalNodeIndex, departureTimeSeconds, arrivalTimeSeconds, tripIndex, canBoard, canUnboard, sequence in trip, lineIndex, blockIndex, canTransferSameLine, minWaitingTimeSeconds
+    std::vector<std::shared_ptr<std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>>> reverseConnections; // tuple: departureNodeIndex, arrivalNodeIndex, departureTimeSeconds, arrivalTimeSeconds, tripIndex, canBoard, canUnboard, sequence in trip, lineIndex, blockIndex, canTransferSameLine, minWaitingTimeSeconds
```
tahini: I think the sorting is very important, otherwise, currently, I don't get the expected data. So I'll replace making this public by a public methods that sets connections, ie will fill the forwardConnections and reverseConnections vectors, sort them and assign them to the trips.

tahini:
#96 is prerequisite and this will be rebased on it later, and add additional tests.

tests/connection_scan_algorithm/csa_test_base.hpp
```diff
@@ -0,0 +1,67 @@
...
+    void setUpPaths();
+    void setUpSchedules();
+
+public:
+    BaseCsaFixtureTests() : params(TrRouting::Parameters()), calculator(TrRouting::Calculator(params)) {}
+    void SetUp();
+
+};
+
+#endif // _CACHE_FETCHER_TEST_H
```
tahini, line 71: Wrong comment at the end

greenscientist (approved):
Just did a quick re-read, but looks good

PR #96, opened 2021-12-10
csa: Refactor/split methods for more modularity and simplicity

tahini:
Preparation work for #94.

These changes extract code that was previously executed in the cache_fetcher or along with cache_fetching, in some cases exposing private fields of the calculator and operating on them without its owning object knowing about it.

The new methods can thus be called independently form different locations. This makes unit tests not only easier, but even possible!

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -68,7 +68,9 @@ extern std::string consoleResetColor;
 
 namespace TrRouting
 {
-  
+  // tuple representing a connection: departureNodeIndex, arrivalNodeIndex, departureTimeSeconds, arrivalTimeSeconds, tripIndex, canBoard, canUnboard, sequence in trip, lineIndex, blockIndex, canTransferSameLine, minWaitingTimeSeconds (-1 to inherit from parameters)
+  using ConnectionTuple = std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>;
+
   class Calculator {
   
   public:
@@ -121,7 +123,7 @@ namespace TrRouting
     int                    updateNetworksFromCache   (Parameters&  params, std::string customPath = "");
     int                    updateSchedulesFromCache  (Parameters&  params, std::string customPath = "");
 
-    int                    setConnections(std::vector<std::shared_ptr<std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>>> connections);
+    int                    setConnections(std::vector<std::shared_ptr<ConnectionTuple>> connections);
     
     int               countStations();
     int               countAgencies();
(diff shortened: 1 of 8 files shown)


connection_scan_algorithm/include/calculator.hpp
```diff
@@ -120,6 +120,8 @@ namespace TrRouting
     int                    updateScenariosFromCache  (Parameters&  params, std::string customPath = "");
     int                    updateNetworksFromCache   (Parameters&  params, std::string customPath = "");
     int                    updateSchedulesFromCache  (Parameters&  params, std::string customPath = "");
+
+    int                    setConnections(std::vector<std::shared_ptr<std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>>> connections);
```
greenscientist: I'm always horrified when I see those tuple!

connection_scan_algorithm/src/initializations.cpp
```diff
@@ -106,6 +106,119 @@ namespace TrRouting
...
 
   }
+
+  int Calculator::setConnections(std::vector<std::shared_ptr<std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>>> connections)
+  {
+    using ConnectionTuple = std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>;
+
+    // Copy the connections to both forward and reverse vectors
+    forwardConnections.clear();
+    forwardConnections.shrink_to_fit();
```
greenscientist: As a small optimisation, you could do the shrink_to_fit after you have done the copy a few lines down. 
That way you won't free memory to just reallocate more right after

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -68,7 +68,9 @@ extern std::string consoleResetColor;
 
 namespace TrRouting
 {
-  
+  // tuple representing a connection: departureNodeIndex, arrivalNodeIndex, departureTimeSeconds, arrivalTimeSeconds, tripIndex, canBoard, canUnboard, sequence in trip, lineIndex, blockIndex, canTransferSameLine, minWaitingTimeSeconds (-1 to inherit from parameters)
+  using ConnectionTuple = std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>;
```
greenscientist, line 72: Maybe typedef this?
oh I guess this is equivalent

greenscientist (approved):
Only minor comments, approving

connection_scan_algorithm/src/initializations.cpp
tahini: Done

connection_scan_algorithm/include/calculator.hpp
tahini: I don't know if it's binary equivalent, but it definitely removes the horrifying tuple from the code!

Issue #95, opened 2021-12-09
CSA: Split the calculator's fields into classes according to their responsibility 


tahini:
The fields of the Calculator class (in `calculator.hpp`) are very numerous, some private, some public, and can be split into classes according to their responsibility:

- [x] Data fields: the agencies, network, services, etc. Singleton for the whole server (if everything is memory) or not (if there's too much data to fit in memory, but we're not there)
- [ ] Calculation fields: node tentative times, trips data, etc. One per alternative calculation, mutable.
- [x] Parameter fields: departure time, access/egress times, etc. One per query, immutable for the whole query

tahini:
parameter fields were completed with PR #113

greenscientist:
Pytting some benchmark data, before I break too much stuff: 
,Forward - no alternatives,Arrival time - no alternatives,Forward - alternatives,Arrival time - alternatives
Short Central Trip (Promenade Masson to Villeray),0.003599,0.004321702,0.330096881,0.438279105
Short Central Trip - outside time range,0.002556168,0.000904083,0.002495002,0.000891709
Long Central Trip (Ahuntsic to Sud-Ouest),0.005211185,0.005928372,1.199193520,1.037634727
Periphery of the transit network (Eastern bout de l'île),0.002771044,0.003067303,0.299603383,0.121376275
Crossing more than half the transit network territory - morning (From PAT to Lasalle),0.012421168,0.011700344,2.647886211,2.400217303
Crossing more than half the transit network territory - noon (From PAT to Lasalle)),0.009350837,0.011089714,2.218660212,3.054860049
Crossing more than half the transit network territory - evening (From PAT to Lasalle),0.009568819,0.013452106,1.964647100,3.200463157
Crossing more than half the transit network territory - night (From PAT to Lasalle),0.005003799,0.007940269,0.206043803,1.366890728
Crossing more than half the transit network territory - outside time range (From PAT to Lasalle),0.000890670,0.000965988,0.001033700,0.000884046
West to east of the transit network (Baie d'urfe to RDP),0.015407004,0.019430594,3.068646565,4.404804738
Central Trip - no alternative (Polytechnique to Villeray),0.003196142,0.003029138,0.005303438,0.005456308
Short Downtown trip,0.002364512,0.002177517,0.013996285,0.013180132

greenscientist:
first of a series to remove the Vector indexes: #166

greenscientist:
Data fields moved to a TransitData object in #196

[...]

PR #98, opened 2022-01-04
csa: Initialize all fields in the calculator constructor

tahini:
It is a rule to initialize all fields of a class in a constructor,
otherwise they may be assigned random values that happen to be in memory
when constructing the object.

The main code path was not affected since there's only one instance of
the Calculator constructed at the beginning of the program. But in unit
tests, it is causing bizarre failures of unrelated tests

### DIFF
--- a/connection_scan_algorithm/src/initializations.cpp
+++ b/connection_scan_algorithm/src/initializations.cpp
@@ -2,11 +2,28 @@
 
 namespace TrRouting
 {
-  
-  Calculator::Calculator(Parameters& theParams) : params(theParams)
+
+  Calculator::Calculator(Parameters& theParams) :
+    projectShortname(""),
+    params(theParams),
+    odTrip(nullptr),
+    origin(nullptr),
+    destination(nullptr),
+    algorithmCalculationTime(CalculationTime()),
+    departureTimeSeconds(0),
+    initialDepartureTimeSeconds(0),
+    arrivalTimeSeconds(-1),
+    maxTimeValue(0),
+    minAccessTravelTime(0),
+    maxEgressTravelTime(0),
+    maxAccessTravelTime(0),
+    minEgressTravelTime(0),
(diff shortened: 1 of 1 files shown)


kaligrafy (approved):

greenscientist (changes requested):
I'd argue that all the objectname() are unnecessary since they will be called automatically by the object creation. 
But the one with non default value sure they are needed. 

Thoughts? Disagreements?

tahini:
> I'd argue that all the objectname() are unnecessary since they will be called automatically by the object creation. But the one with non default value sure they are needed.
> 
> Thoughts? Disagreements?

@greenscientist Your expertise in c++ will be helpful here: the original problem I had that made me do this patch, was that in the unit test, the constructor was called and reassigned memory for this object and probably at the location of the odTrip pointer, there was junk data and the pointer had a value, which caused errors in tests using the object. Will all non-pointer fields be automatically assigned with default constructor? Or will be they simply point to some memory location that may contain data that will make it think the vector is not empty? I've been reading this https://isocpp.org/wiki/faq/ctors#init-lists before doing this patch. It doesn't mention the case of 50 fields in a class though... ;-)

greenscientist:
So anythying that as a constructor will get initialized automatically. 
Other things like pointers, int, floats, etc. Need a manual init. 

For example: 

#include <iostream>

class B {
public:
  B() {
    std::cout << "B INIT" << std::endl;
  }
};

class A {
public:
  A() {
    std::cout << "A INIT" << std::endl;
  }
  B test;
};

int main() {
  A toto;
}

./a.out 
B INIT
A INIT

kaligrafy (approved):

greenscientist (approved):
Ok

One thing that the early version of this diff highligts is that there's way too many data objects in thie Calculator class!

tahini:
> One thing that the early version of this diff highligts is that there's way too many data objects in thie Calculator class!

Issue #95 tracks this. These PRs are preparation for the "parameters field" task, but the rest should come soon too.

PR #101, opened 2022-01-06
osrm_fetcher: Pass the max travel time and walking speed as parameter

### LINKED ISSUE #97: The max_egress_travel_time_seconds is never considered
It is possible to set a parameter called max_egress_travel_time_seconds when doing a query, but that parameter is never considered.

The osrm_fetcher compares access time (params.maxAccessWalkingTravelTimeSeconds), but params.maxEgressWalkingTravelTimeSeconds is never used in the code.

2 options:

1- Fix it and send to the osrm_fetch the maximal value to consider, access or egress time, depending on the context.

2- ~~Remove the max egress travel time parameter and consider symetry between access and egress.~~

@kaligrafy  which option is to be implemented? And long live unit tests! :p

tahini:
Fixes #97

The Parameters object is the server parameters' main object, that
contains the connection information. The speed and max travel time will
soon be extracted out of this class. Also, depending on the calling
context, the max travel time may not be the access time, but the egress
or any other time. This makes these methods more flexible.

Unit test for max egress time is brought back as it now works.

### DIFF
--- a/include/osrm_fetcher.hpp
+++ b/include/osrm_fetcher.hpp
@@ -31,14 +31,14 @@ namespace TrRouting
     {
     }
 
-    static std::vector<std::tuple<int, int, int>> getAccessibleNodesFootpathsFromPoint(const Point point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, bool reversed = false);
+    static std::vector<std::tuple<int, int, int>> getAccessibleNodesFootpathsFromPoint(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, int maxWalkingTravelTime, float walkingSpeedMetersPerSecond, bool reversed = false);
 
   protected:
-    static std::vector<std::tuple<int, int, int>> getNodesFromBirdDistance(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, Parameters &params);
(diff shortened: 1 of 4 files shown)


kaligrafy (approved):

include/osrm_fetcher.hpp
```diff
@@ -31,14 +31,14 @@ namespace TrRouting
...
 
   protected:
-    static std::vector<std::tuple<int, int, int>> getNodesFromBirdDistance(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, Parameters &params);
-    static std::vector<std::tuple<int, int, int>> getNodesFromOsrm(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, bool reversed);
+    static std::vector<std::tuple<int, int, int>> getNodesFromBirdDistance(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, Parameters &params, int maxWalkingTravelTime, float walkingSpeedMetersPerSecond);
+    static std::vector<std::tuple<int, int, int>> getNodesFromOsrm(const Point &point, const std::vector<std::unique_ptr<Node>> &nodes, std::string mode, Parameters &params, int maxWalkingTravelTime, float walkingSpeedMetersPerSecond, bool reversed);
 
     static std::tuple<float, float> calculateLengthOfOneDegree(const Point &point);
-    static float calculateMaxDistanceSquared(Parameters &params);
+    static float calculateMaxDistanceSquared(int maxWalkingTravelTime, float walkingSpeed);
```
greenscientist, line 41: Good, reducing the scope of information sent to a function!

greenscientist (approved):
Good job here, we need more of those! :)

PR #102, opened 2022-01-07
Extract server and data validation code to their own functions

tahini:
The data initialization function now returns the status of the data, so that instead of checking the number of each data at the beginning of a request, it is done once. The response initialization is done in its own function so it can be reused by other endpoints.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -74,6 +74,28 @@ namespace TrRouting
   class Calculator {
   
   public:
+
+    enum class DataStatus 
+    {
+      // Data is ready for query
+      READY = 0,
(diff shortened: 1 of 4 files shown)


kaligrafy (approved):

greenscientist (approved):
About the first commit, it would be nice, eventually to have an actual Response class and not just a string, so we can track its state properly. 

For the second. encapsulating the error code in an ENUM is a code first step. I kinda feel we should have a DataError and DataOk "meta class", or a function in the object to quickly check "isError()" instead of any to == to DATA_READY, but again, good improvement and good enough. We'll probably do more in-depth refactoring

tahini:
Thanks for the review. There will indeed be more refactoring eventually, when there will be more endpoints. For now, it provides a quick response path.

PR #103, opened 2022-01-07
Add a configure flag to enable benchmarks

### LINKED ISSUE #92: Move the CSA benchmark to a new make target
Now, the benchmark is run using make check, the same command as for unit tests.

Since benchmarks are meant to run for a long time (if we really want to make valid benchmarks...), it should be its own command: `make bench` for example

tahini:
The --enable-benchmark flag when running ./configure will enable the
benchmark compilation and execution when running `make check`. By
default, the benchmarks are disabled.

Fixes #92. It's not a different target, but at least the benchmarks
won't run when running unit tests by default.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -46,6 +46,17 @@ LDFLAGS+=" $BOOST_LDFLAGS $BOOST_REGEX_LIB $BOOST_FILESYSTEM_LIB $BOOST_IOSTREAM
 LDFLAGS+=" $CURSES_LIBS"
 LDFLAGS+=" $CAPNP_LIBS"
 
+# Do not enable benchmark in the make check by default
+AC_ARG_ENABLE([benchmark],
+    [  --enable-benchmark      Benchmarks will be run when running make check],
+    [case "${enableval}" in
+        yes) benchmark=true ;;
+        no)  benchmark=false ;;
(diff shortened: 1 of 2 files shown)


greenscientist (approved):
Fun with M4 I see... :)
Yep, that's a good solution. 
But you have a bit of commented code.

configure.ac
```diff
@@ -46,6 +46,20 @@ LDFLAGS+=" $BOOST_LDFLAGS $BOOST_REGEX_LIB $BOOST_FILESYSTEM_LIB $BOOST_IOSTREAM
...
+    [case "${enableval}" in
+        yes) benchmark=true ;;
+        no)  benchmark=false ;;
+        *) AC_MSG_ERROR([bad value ${enableval} for --enable-benchmark]) ;;
+    esac],[benchmark=false])
+AS_IF([test x"$benchmark" = xfalse], [AC_MSG_NOTICE([Benchmarks are disabled. Enable them by adding the --enable-benchmark flag to configure.])])
+AM_CONDITIONAL([RUN_BENCHMARK], [test x$benchmark = xtrue])
+
+
+# AM_CONDITIONAL([COND_BENCHMARK], [test "$enable_benchmark" = yes])
```
greenscientist: Commented code

PR #104, opened 2022-01-11
Rename the Parameters class to ServerParameters

tahini:
This will allow to eventually have another class for RouteParameters
where the part of the parameters that are specific for a single route query
will be moved.

Note that the ServerParameters class still contains a lot of non-server
parameters.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -100,7 +100,7 @@ namespace TrRouting
     std::map<std::string, int> benchmarking;
     std::string projectShortname;
     
-    Calculator(Parameters& theParams);
+    Calculator(ServerParameters& theParams);
 
     /**
      * Prepare the data for the calculator and validate that there are at least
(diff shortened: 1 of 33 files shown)


greenscientist (commented):
I'm not sure that Server is an appropropriate name. CalculationParam ? InstanceParam ? 
There's a few thing yes about a deamon instance, but many other things. Thoughts?

tahini:
> I'm not sure that Server is an appropropriate name. CalculationParam ? InstanceParam ?

Maybe InstanceParam, but indeed there's currently more Calculation parameters than instance params in this class, but since I was thinking to make a refactoring to extract the calculation params, the CalculationParam would be a new class in that later PR. But technically, I could also extract the Instance Parameters from the Calculation parameters, but it's just how you see it :p 

Maybe it's too early to rename this class then...

greenscientist:
> 
> Maybe it's too early to rename this class then...

Maybe easier to keep this one, and chip pieces of it and then name those small chunk

greenscientist:
But in the current context, since the deamon is everything. Saying it's a ServerParam is also true. I'll accept the change and let you decide if you prefer to abandon the diff or keep it

kaligrafy (approved):

tahini:
I'll abandon, this class will be either removed or simply renamed when only one type of params remain in it.

PR #105, opened 2022-01-11
Turn the CSA benchmark into a parameterized test

tahini:
This allows to specify a list of parameters for each calculation to make
and run benchmark for each case.

For each test case, the data is calculated with or without alternatives
and using the time as arrival or departure time.

The results are written to 2 csv files:

* `benchmarkResults.csv` contains a line for each test case and the
results for each combination of alternatives or not and arrival or
departure time (so 4 results).
* `benchmarkResultsDetailed.csv` contains a line per combination with
the individual results for each calculation.

### DIFF
--- a/tests/benchmark_csa/benchmark_CSA_test.cpp
+++ b/tests/benchmark_csa/benchmark_CSA_test.cpp
@@ -16,7 +16,8 @@
 #include <bits/stdc++.h>
 #include <chrono>
 #include <iostream>
-#include <fstream> 
+#include <fstream>
+#include <ctime>
 
 #include "toolbox.hpp"
 #include "gtfs_fetcher.hpp"
@@ -28,19 +29,54 @@
 #include "calculator.hpp"
 #include "program_options.hpp"
 #include "benchmark_CSA_test.hpp"
+#include "constants.hpp"
 
 using namespace TrRouting;
 
+const int NB_ITER = 30;
+
+// Global test suite variables, they should not be reset for each test
+Parameters algorithmParams;
+Calculator* calculator;
+std::ofstream benchmarkResultsFile;
+std::ofstream benchmarkDetailedResultsFile;
+
+/**
+ * The benchmarks require cache data to be available in the
+ * tests/benchmark_csa/cache/demo_transition directory. The
+ * cache data to match the benchmark can be found here:
+ * https://nuage.facil.services/s/Gs7bwfEGaBjHCLg Simply unzip in the
+ * tests/benchmark_csa/cache directory. It corresponds to the
+ * STM's fall 2018 18S_S service (Société des Transports de Montréal).
+ * */
 class BenchmarkCSATests : public ConstantBenchmarkCSATests
 {
 
 protected:
-  int numberOfRequests = 0;
 
-  Parameters setupAlgorithmParams()
+  inline static const boost::uuids::string_generator uuidGenerator;
+  // String uuids of various objects
+  inline static const boost::uuids::uuid scenarioUuid = uuidGenerator("ed42d920-0349-4f64-8590-4698056c2734");
+
+
+  static bool updateCalculatorFromCache(Calculator *calculator)
   {
-    Parameters algorithmParams;
+    std::cout << "Preparing calculator..." << std::endl;
+    Calculator::DataStatus dataStatus = calculator->prepare();
 
+    if (dataStatus != Calculator::DataStatus::READY)
+    {
+      std::cout << "Something is wrong with the calculator. Data status: " << (int) dataStatus << std::endl;
+      return false;
+    }
+    return true;
+  }
+
+public:
+
+  // Initialize calculator and parameters. Open the result files and add headers
+  static void SetUpTestSuite()
+  {
     algorithmParams.projectShortname = "demo_transition";
     algorithmParams.cacheDirectoryPath = "cache";
     algorithmParams.dataFetcherShortname = "cache";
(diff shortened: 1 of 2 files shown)


kaligrafy (approved):

tahini:
In a single run of the benchmarks with 30 iteration per benchmark, the standard deviation is below 15% for >90% of the cases. Between runs, the results vary, but the benchmarks are run on my working machine, while I work. I should try when the machine is practically idle and see if there is more stability.

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -112,87 +122,124 @@ class BenchmarkCSATests : public ConstantBenchmarkCSATests
```
greenscientist: Should probably test here that the calcultator pointer you receive is not null. 
Alternatively, you could consider passing a reference

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -112,87 +122,124 @@ class BenchmarkCSATests : public ConstantBenchmarkCSATests
...
-      {
-        return true;
-      }
+      return true;
     }
     return false;
   }
 
-  void benchmarkCurrentParams(TrRouting::Calculator *calculator)
+  void benchmarkCurrentParams(std::vector<std::string> *parametersWithValues, bool expectResult)
```
greenscientist: Just wondering if NB_ITER should be a paremeter of the function. (And the call could pass the constant)

greenscientist (commented):
Looks ok, I worry a little bit about the calculator pointer, but we are still in C++, still can happen...

tests/benchmark_csa/benchmark_CSA_test.cpp
tahini: For now, all queries take about the same time, or at least nothing beyond a few seconds, so having different iterations count isn't really relevant. Unless you mean as a parameter to test test app, like `NB_ITER=100 make check`, to run a different number of iterations without having to compile again?

greenscientist: It would be to eventually maybe do something like that. But not critical at the moment we can add it later

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -28,19 +28,54 @@
...
+#include "constants.hpp"
 
 using namespace TrRouting;
 
+const int NB_ITER = 30;
+
+// Global test suite variables, they should not be reset for each test
+Parameters algorithmParams;
+Calculator* calculator;
+std::ofstream benchmarkResultsFile("benchmarkResults.csv");
```
greenscientist: Just to be really picky, you can add a TODO and an issue if you want. 

We should have a timestamp or something like that on the filename. So that you don't always overwrite previous benchmark.

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -58,145 +93,160 @@ class BenchmarkCSATests : public ConstantBenchmarkCSATests
...
-    // ASSERT_EQ(true, false);
-    ASSERT_EQ(true, true);
-  }
-}
\ No newline at end of file
+// Values are: (description, lon, lat orig, lon, lat dest, time, whether results are expected)
+INSTANTIATE_TEST_SUITE_P(
+  BenchmarkSimpleQueries, BenchmarkCSATests,
+  testing::Values(
+    std::make_tuple("Short Central Trip", -73.577867, 45.552398, -73.622594, 45.542273, 8 * 60 * 60, true),
```
greenscientist: DO we have some idea what those coordinate are?

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -58,145 +93,160 @@ class BenchmarkCSATests : public ConstantBenchmarkCSATests
...
+    std::make_tuple("Short Central Trip - outside time range", -73.577867, 45.552398, -73.622594, 45.542273, 4 * 60 * 60, false),
+    std::make_tuple("Long Central Trip", -73.64579, 45.576736, -73.563364, 45.477621, 12 * 60 * 60, true),
+    std::make_tuple("Eastern bout de l'île Trip", -73.538513, 45.670863, -73.49693, 45.662265, 12 * 60 * 60, true),
+    std::make_tuple("From PAT to Lasalle - morning", -73.487088, 45.693797, -73.636331, 45.421016, 8 * 60 * 60 + 30 * 60, true),
```
greenscientist: Those name will not make any sense for anyone outside of Montreal. Probably need more comments

tests/benchmark_csa/benchmark_CSA_test.cpp
tahini: Just arbitrary clicks around the map, to give trips of various length and various amount of alternatives.

tests/benchmark_csa/benchmark_CSA_test.cpp
tahini: I'll rephrase with something relative to the network, but then it won't make sense to us anymore...

tests/benchmark_csa/benchmark_CSA_test.cpp
tahini: Good idea

tests/benchmark_csa/benchmark_CSA_test.cpp
greenscientist: just make sure there's comment that makes sense to us and to the rest of the world

tahini:
I updated the titles of the benchmarks

greenscientist (approved):

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -16,7 +16,8 @@
 #include <bits/stdc++.h>
 #include <chrono>
 #include <iostream>
-#include <fstream> 
+#include <fstream>
+#include <time.h>
```
greenscientist: I think the proper include is #include <ctime>

tahini: done, time.h is c and ctime is c++, so ctime it is.

PR #106, opened 2022-01-11
csa: Remove the only_*/except_* parameters from query

tahini:
Those parameters are not used now, the scenario can define include and
exclude object lists.


kaligrafy (approved):

greenscientist (approved):
ok sure

Issue #100, opened 2022-01-06
Switch lat,lon string in queries to lon,lat to follow standard x,y and geojson format


kaligrafy:
Mega breaking change, we must find all references to trRouting calls everywhere!

tahini:
I would argue to leave intact the current endpoint, so as not to do this mega breaking change. But the new endpoints will be in lon/lat order.

PR #107, opened 2022-01-13
Document a new endpoint for origin/destination routing queries

tahini:
This commit is the start of v2 of the trRouting API, where each query
will give access to a single functionality and parameters are
well-defined and for each query, all of the parameters can be set.

This endpoint returns a single transit routing calculation for the
origin/destination in parameter. It is a subset of the route/v1/transit
query, so return types are kept as they are in the original endpoint.

Note that this endpoint reverses the coordinates order of the
parameters. Origin and destination are now lon/lat, instead of lat/lon,
as mentioned in #100.

### DIFF
--- a/docs/API.yml
+++ b/docs/API.yml
@@ -105,7 +105,11 @@ paths:
             default: 0
           required: false
           # TODO: This will become 2 different endpoints since the return values are so much different.
-          description: If set to 0, the query returns the routes between origin and destination. If set to 1, the query will return all the nodes accessible with the given parameter. Only the origin is required for this query.
+          description: |
+            If set to 0, the query returns the routes between origin and destination.
+            If set to 1, the query will return all the nodes accessible with the given parameter.
+            Only the origin is required for this query that will calculate all nodes accessible
+            from this point. There can also be only the destination for all nodes to this point.
       # The parameters above are those used in the Transition queries, already ported to typescript
       # Still undocumented parameters. See if they are all required and remove any reference to them if they are not. Parameters on the same line are in the same if condition in the code, so identical
       # Some parameters differ only with the unit, consider using a single parameter, and if really required, add a parameter that contains the units. Otherwise, one can specify 2 such parameters, which one to take?
--- a/docs/APIv2.yml
+++ b/docs/APIv2.yml
@@ -1,0 +1,449 @@
+openapi: '3.0.2'
+info:
+  title: TrRouting
+  version: '2.0-beta'
+servers:
+  - url: http://localhost:4000
+
+paths:
+  /versions:
+    get:
+      description: Get the versions of the API supported by this inVehicleDistanceMeters
+      responses:
+        '200':
+          description: Return the current and supported versions of the API
+          content:
+            application/json:
+              schema:
+                type: object
+                properties:
+                  current:
+                    type: string
+                    description: Current (latest) version of the API
+                  supported:
+                    type: array
+                    items:
+                      type: string
+                    description: All supported versions of the API
+
+  /v2/route:
+    get:
+      description: Calculate transit route results for a given origin/destination pair
+      parameters:
+        - in: query
+          name: origin
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: true
+          description: Comma-separated longitude/latitude coordinates of the origin, in the WSG84 coordinates system
+        - in: query
+          name: destination
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: true
+          description: Comma-separated longitude/latitude coordinates of the destination, in the WSG84 coordinates system
+        - in: query
+          name: scenario_id
+          schema:
+            type: string
+          required: true
+          # TODO: Scenario is very transition centric. Can't trRouting be used with a GTFS or some other type, without a scenario
+          description: ID of the scenario to query. A scenario defines the services, agencies and lines to use for this transit calculation
+        - in: query
+          name: time_of_trip
+          schema:
+            type: integer
+          required: true
+          description: |
+            time of the trip, in seconds since midnight.
+            The time_type field will determine if it represents a departure or arrival time.
+            There is no timezone associated with the time, trRouting is timezone agnostic as a scenario
+            typically covers a single timezone and the 0 is the midnight in the agency of that scenario.
+        - in: query
+          name: time_type
+          schema:
+            type: integer
+            enum:
+              - 0
+              - 1
+          required: false
+          description: The type of the time_of_trip. 0 means it is the departure time; 1 means arrival time. Defaults to 0.
+        - in: query
+          name: alternatives
+          schema:
+            type: boolean
+          required: false
+          description: Whether the results should return various alternatives if available or just a single result. Defaults to false, no alternatives
+        - in: query
+          name: min_waiting_time
+          schema:
+            type: integer
+          required: false
+          description: "The minimum time to wait at a stop/station, in seconds, to cope with uncertainties in the vehicle arrival times. Suggested value: 180"
+        - in: query
+          name: max_access_travel_time
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, to reach the first stop/station in the trip"
+        - in: query
+          name: max_egress_travel_time
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, from the last stop/station, to reach the destination"
+        - in: query
+          name: max_transfer_travel_time
+          schema:
+            type: integer
+          required: false
+          description: "Maximum time, in seconds, for each transfer between stop/station during the trip"
+        - in: query
+          name: max_travel_time
+          schema:
+            type: integer
+          required: false
+          description: The maximum total travel time between origin and destination, including access, transfer and egress times
+        - in: query
+          name: max_first_waiting_time
+          schema:
+            type: integer
+          required: false
+          description: The maximum time, in seconds, one can wait at first stop/station to consider this trip valid
+      responses:
+        '200':
+          description: Successful query, but may not have returned a routing result
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/data_error'
+                  - $ref: '#/components/schemas/NoRoutingFound'
+                  - $ref: '#/components/schemas/success'
+                discriminator:
+                  propertyName: status
+        '400':
+          description: Query parameters are invalid
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/query_error'
+
+components:
+  schemas:
+    statusResponse:
+      type: object
+      required:
+        - status
+      properties:
+        status:
+          type: string
+          enum: [success, no_routing_found, data_error, query_error]
+      discriminator:
+        propertyName: status
+        mapping:
+          success: success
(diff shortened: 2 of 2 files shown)


kaligrafy (approved):

docs/API.yml
```diff
@@ -169,7 +173,110 @@ paths:
...
+          description: Comma-separated longitude/latitude coordinates of the origin, in the WSG84 coordinates system
+        - in: query
+          name: destination
+          schema:
+            type: string
+            pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          required: true
+          description: Comma-separated longitude/latitude coordinates of the destination, in the WSG84 coordinates system
+        - in: query
+          name: scenario_uuid
```
greenscientist: I wonder if we should call it uuid or just scenario_id
I can envision that we could use with something else than transition, but still want to identifiy it's data source somehow. But have an UUID all the time seems a bit heavy. 
Thoughts?

docs/API.yml
```diff
@@ -169,7 +173,110 @@ paths:
...
+          schema:
+            type: string
+          required: true
+          # TODO: Scenario is very transition centric. Can't trRouting be used with a GTFS or some other type, without a scenario
+          description: UUID of the scenario to query. A scenario defines the services, agencies and lines to use for this transit calculation
```
greenscientist: We might want to specify that the scenario also define the timezone??
(I'm thinking about the time parameters later)

docs/API.yml
```diff
@@ -169,7 +173,110 @@ paths:
...
+        - in: query
+          name: time_of_trip
+          schema:
+            type: integer
+          required: true
+          description: time of the trip, in seconds since midnight. The reverse field will determine if it represents a departure or arrival time.
+        - in: query
+          name: is_arrival
+          schema:
+            type: integer
```
greenscientist: with not bool?

greenscientist: Reading the description, this feel it should be named some thing like time type and be an enum

greenscientist (changes requested):
Some questions, 
requesting change to allow for discussion

docs/API.yml
tahini: Indeed, who knows what the future has in store for trRouting, scenario_id is nicer

docs/API.yml
tahini: I don't know what other kind of time it can be. In the code, we use `forwardCalculation`, so it can be a boolean field called forward. I forgot boolean was allowed with openAPI, I just didn't use it before before it's too restrictive and the "boolean" fields were taking 1 as value instead of true, but here, it's a new endpoint, so might as well make it a boolean.

greenscientist: I always have 3 choice when I use a routing app, let me check.

tahini: Usually the third choice is "right now", no?

docs/API.yml
```diff
@@ -169,7 +173,110 @@ paths:
...
+      responses:
+        '200':
+          description: Successful query, but may not have returned a routing result
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/data_error'
+                  - $ref: '#/components/schemas/no_routing_found'
+                  - $ref: '#/components/schemas/success'
```
tahini: I'm keeping the responses as is, but the response has a lot of duplicate fields, like times in minutes AND seconds, sometimes even as human readable string. Should we keep only one of each, the field in seconds, from which, when required, we can calculate the other units? Is it trRouting's role to make those conversions or should it be the caller?

@kaligrafy what would be the impact on you to receive only the fields in seconds only? On the network, it should reduce the size of the response's body by close to the half. On the computation time, it's probably negligible.

kaligrafy: no problem getting the seconds only. Only old tools used the minutes.

docs/API.yml
greenscientist: True. 
But now I've checked google maps and they have a Last Available option. 
So boolean or enum (we can just have 0, 1 for now but still allow for easy extention. 
Not attached to either solution.

docs/API.yml
greenscientist: Yeah second only. Since we send string anyway, it does  not impact much on data size. 
@kaligrafy I imagine we won't need millisecond precision on that kind of calculatioN?

docs/API.yml
tahini: I named it time_type and left it an int in case we need to add options eventually, to not break the API

docs/API.yml
```diff
@@ -169,7 +173,111 @@ paths:
...
+      responses:
+        '200':
+          description: Successful query, but may not have returned a routing result
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: '#/components/schemas/data_error'
+                  - $ref: '#/components/schemas/NoRoutingFoundV2'
+                  - $ref: '#/components/schemas/successV2'
```
tahini: I added a v2 for success and noRoutingFound. I had to duplicate most of the return types to remove duplication and change the lat/lon to lon/lat formats.

kaligrafy (approved):

docs/API.yml
kaligrafy: I assume there will be another url route for routing from/to all stops for accessibility maps?

tahini: Yes, whenever another set of parameters is required, there will be another route. This avoids the trouble of complex parameter validation: if X is 0, then Y and Z ar required, but if X is 1, then either Y OR Z is required and what if both are set should we return a warning.... No, here, each param is either required or optional and if optional there is only one  default value for it, it's easy to validate, no possible error and if wrong, we return a query_error

docs/API.yml
```diff
@@ -556,4 +687,277 @@ components:
...
+      allOf:
+        - $ref: '#/components/schemas/baseRouteResponseV2'
+        - type: object
+          properties:
+            departureTime:
+              type: string
+              description: Optimized departure time in seconds since midnight
+            initialDepartureTime:
+              type: string
+              description: Departure time as specified in the request, in seconds since midnight, or -1 if the departure time was not specified
```
greenscientist: I'm just curious, how to we handle change of day ?
Example: initial departure time is 23h55, but the actual departure is 00h05. Do we have an indication that it's the next day?

greenscientist: Also, do we need to replicate the intiialDepartureTime? Can we assume that the user knows this already?

docs/API.yml
```diff
@@ -556,4 +687,277 @@ components:
...
+            initialLostTimeAtDeparture:
+              type: number
```
greenscientist: Seems like this is covered by having departureTime and initialDepartureTime

docs/API.yml
```diff
@@ -556,4 +687,277 @@ components:
...
+        origin:
+          type: string
+          pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          description: Comma-separated longitude/latitude coordinates of the origin, in the WSG84 coordinates system
+        destination:
+          type: string
+          pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+          description: Comma-separated longitude/latitude coordinates of the destination, in the WSG84 coordinates system
+
+    routeResponseV2:
```
greenscientist: I'm already thinking of a good V3 for this type of response... :)

docs/API.yml
```diff
@@ -556,4 +687,277 @@ components:
...
+      - $ref: '#/components/schemas/tripStepV2'
+      - type: object
+        properties:
+          agencyAcronym:
+            type: string
+            description: Acronym of the agency serving the route being boarded/unboarded
+          agencyName:
+            type: string
+            description: Name of the agency serving the route being boarded/unboarded
+          agencyUuid:
```
greenscientist: Why do we return the Name if we return the UUID?

greenscientist (commented):
few more comments

docs/API.yml
tahini: I would keep the initialDepartureTime, it's an important parameter query and reminding it is very helpful. As for the change of day, in transit, days have more than 24 hours, so 00h05 would be 24h05 and 1h00 would be 25h00. But what if a trip is done partly with a night bus until 28h00, but ends with a day bus that starts at 4h30? @kaligrafy  would trRouting work for those trips?

docs/API.yml
tahini: > I'm already thinking of a good V3 for this type of response... :)

Anything to say?

tahini:
In general, I see the response as a standalone json object that can be dumped somewhere and contains all the required information, some of which can be computed from the fields, and including the important query parameters (origin, destination, time and probably time_type to know if it was arrival or departure that was requested). 

That's why all the redundant unit fields are gone, and we can also remote the initialLostDepartureTime as it can be computed from other fields and is not required if it's an arrival time anyway.

As for the agency, name and uuid, I would keep both, as the uuid is important for a tool like transition to retrieve the lines and paths and get geojson of those segments, but it is meaningless for someone just reading a dump of the response, while the name tells right away which line/path were taken by the route. What do you think?

tahini:
Actually, the API version is in the API.yml file, so I think I should create a new file: APIv2.yml to put this new endpoint, and let the version be 2.0-beta. Then we can go ahead without feeling too committed to whatever is here and we can improve as all the new endpoints are added and their implementations are developped. How does that sound?

tahini:
The v2 API is officially in beta, so we won't have to feel too bad about changing it in a near future if we realize something better can be done, but we can go ahead with this at least. Also note that adding functionalities to a route (like an eventual cost function parameters) is not considered an API break and does not require major update

docs/API.yml
greenscientist: Something more similar to the trip steps. Right now this responce object only assume  to walk and take bus/train. But we can combine other transport. (People in the suburbs takes their care to the train stop for example). So instead of a fixed list of parameters, we should combine it with the appropriate objects if needed

greenscientist:
> As for the agency, name and uuid, I would keep both, as the uuid is important for a tool like transition to retrieve the lines and paths and get geojson of those segments, but it is meaningless for someone just reading a dump of the response, while the name tells right away which line/path were taken by the route. What do you think?

Maybe we should have 2 query time then, one for machine consumption and one for more human needs?

greenscientist:
I guess the file split is ok if you say so :)

Maybe we should add a query to ask the server which queries version is support...

docs/API.yml
tahini: You're pointing to a wide opened Pandora's box... Is it even trRouting's role to calculate the walking time? trRouting has its own way of calculating the walking time: either using osrm with a 'walking' profile (in our setup, get the wrong osrm port and you've got totally different results!), or use bird distance if no osrm server is available.

The walking steps results shows the times of those steps, _as calculated by trRouting_. But the tool that receives it (eg transition) and wants to display it as a geojson needs to calculate it again from the coordinates and that tools can very well have different results!

trRouting is the door for the connection scan algorithm (and eventually others) that find transit trips. Is multimodality its purpose? Maybe it is, but not in its current state. And if it does return trips for other modes, it should probably return the entire information to avoid the calling tool calculating again the geography with potentially different data.

But all this can be further API enhancements on v2, I don't think the proposition here is painting us in any corner.

tahini:
I added the /versions endpoint

greenscientist:
I'll approve this as is. I think we can do a lot more improvement to this API, but this probably imply more philosophical changes, that can go into a V3. This V2 is more about simplifying the API (I kinda think we might have too many fields still, but it's an improvement)

greenscientist (approved):

PR #108, opened 2022-01-13
Add a RouteParameters class to store route query specific parameters

tahini:
This class contains the parameters passed to a single routing query:
origin/destination, max/min times for waiting, access, egress, transfer,
etc and scenario data.

Add a factory method to create a parameters object from a parameters
vector.

Add unit tests for this factory method.

### DIFF
--- a/connection_scan_algorithm/src/route_parameters.cpp
+++ b/connection_scan_algorithm/src/route_parameters.cpp
@@ -1,0 +1,280 @@
+#include "parameters.hpp"
+
+namespace TrRouting
+{
+  static int DEFAULT_MIN_WAITING_TIME = 3 * 60;
+  static int DEFAULT_MAX_TOTAL_TIME = MAX_INT;
+  static int DEFAULT_MAX_ACCESS_TRAVEL_TIME = 20 * 60;
+  static int DEFAULT_MAX_EGRESS_TRAVEL_TIME = 20 * 60;
+  static int DEFAULT_MAX_TRANSFER_TRAVEL_TIME = 20 * 60;
+  static int DEFAULT_FIRST_WAITING_TIME = 30 * 60;
+
+  RouteParameters::RouteParameters(std::unique_ptr<Point> orig_,
+    std::unique_ptr<Point> dest_,
+    Scenario& scenario_,
+    int _timeOfTrip,
+    int minWaitingTime,
+    int maxTotalTime,
+    int maxAccessTime,
+    int maxEgressTime,
+    int maxTransferTime,
+    int maxFirstWaitingTime,
+    bool alt,
+    bool forward) :
+        origin(std::move(orig_)),
+        destination(std::move(dest_)),
+        scenario(scenario_),
+        timeOfTrip(_timeOfTrip),
+        minWaitingTimeSeconds(minWaitingTime),
+        maxTotalTravelTimeSeconds(maxTotalTime),
+        maxAccessWalkingTravelTimeSeconds(maxAccessTime),
+        maxEgressWalkingTravelTimeSeconds(maxEgressTime),
+        maxTransferWalkingTravelTimeSeconds(maxTransferTime),
+        maxFirstWaitingTimeSeconds(maxFirstWaitingTime),
+        withAlternatives(alt),
+        forwardCalculation(forward)
+  {
+    scenarioUuid = scenario.uuid;
+    onlyServicesIdx = scenario.servicesIdx;
+    onlyLinesIdx = scenario.onlyLinesIdx;
+    onlyAgenciesIdx = scenario.onlyAgenciesIdx;
+    onlyNodesIdx = scenario.onlyNodesIdx;
+    onlyModesIdx = scenario.onlyModesIdx;
+    exceptLinesIdx = scenario.exceptLinesIdx;
+    exceptAgenciesIdx = scenario.exceptAgenciesIdx;
+    exceptNodesIdx = scenario.exceptNodesIdx;
+    exceptModesIdx = scenario.exceptModesIdx;
+  }
+
+  RouteParameters::RouteParameters(const RouteParameters& routeParams):
+    origin(std::make_unique<Point>(routeParams.origin.get()->latitude, routeParams.origin.get()->longitude)),
+    destination(std::make_unique<Point>(routeParams.destination.get()->latitude, routeParams.destination.get()->longitude)),
+    scenario(routeParams.scenario),
+    timeOfTrip(routeParams.timeOfTrip),
+    minWaitingTimeSeconds(routeParams.minWaitingTimeSeconds),
+    maxTotalTravelTimeSeconds(routeParams.maxTotalTravelTimeSeconds),
+    maxAccessWalkingTravelTimeSeconds(routeParams.maxAccessWalkingTravelTimeSeconds),
+    maxEgressWalkingTravelTimeSeconds(routeParams.maxEgressWalkingTravelTimeSeconds),
+    maxTransferWalkingTravelTimeSeconds(routeParams.maxTransferWalkingTravelTimeSeconds),
+    maxFirstWaitingTimeSeconds(routeParams.maxFirstWaitingTimeSeconds),
+    withAlternatives(routeParams.withAlternatives),
+    forwardCalculation(routeParams.forwardCalculation),
+    scenarioUuid(routeParams.scenarioUuid),
+    onlyServicesIdx(routeParams.onlyServicesIdx),
+    onlyLinesIdx(routeParams.onlyLinesIdx),
+    onlyAgenciesIdx(routeParams.onlyAgenciesIdx),
+    onlyNodesIdx(routeParams.onlyNodesIdx),
+    onlyModesIdx(routeParams.onlyModesIdx),
+    exceptLinesIdx(routeParams.exceptLinesIdx),
+    exceptAgenciesIdx(routeParams.exceptAgenciesIdx),
+    exceptNodesIdx(routeParams.exceptNodesIdx),
+    exceptModesIdx(routeParams.exceptModesIdx)
+  {
+  }
+
+  static int getIntegerValue(std::string strValue) {
+    try
+    {
+      return std::stoi(strValue);
+    }
+    catch (...)
+    {
+      // Throwing an error, but do we want to fallback to default and just ignore this one?
+      throw ParameterException(ParameterException::Type::INVALID_NUMERICAL_DATA);
+    }
+  }
+
+  RouteParameters RouteParameters::createRouteODParameter(std::vector<std::pair<std::string, std::string>> &parameters, std::map<boost::uuids::uuid, int> &scenarioIndexesByUuid, std::vector<std::unique_ptr<Scenario>> &scenarios)
+  {
+
+    boost::uuids::string_generator uuidGenerator;
+
+    Scenario * scenario = nullptr;
+    boost::optional<boost::uuids::uuid> scenarioUuid = boost::none;
+    boost::uuids::uuid originNodeUuid;
+    boost::uuids::uuid destinationNodeUuid;
+
+    // Initialize default values
+    int timeOfTrip = -1;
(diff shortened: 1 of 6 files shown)


tahini:
This is the parameter parsing part of the new endpoint, as described in PR #107

kaligrafy (approved):

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
+#include "parameters.hpp"
+
+namespace TrRouting
+{
+
+  RouteParameters::RouteParameters(std::unique_ptr<Point> orig_,
+    std::unique_ptr<Point> dest_,
+    Scenario* scenario_,
```
greenscientist: Can this be a reference instead? Can the pointer ever be null?

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
...
+        minWaitingTimeSeconds(minWaitingTime),
+        maxTotalTravelTimeSeconds(maxTotalTime),
+        maxAccessWalkingTravelTimeSeconds(maxAccessTime),
+        maxEgressWalkingTravelTimeSeconds(maxEgressTime),
+        maxTransferWalkingTravelTimeSeconds(maxTransferTime),
+        maxFirstWaitingTimeSeconds(maxFirstWaitingTime),
+        alternatives(alt),
+        forwardCalculation(forward)
+  {
+    scenarioUuid = scenario->uuid;
```
greenscientist: Answering my earlier question the pointer cannot be null, since we do not check for that case here.

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
...
+    minWaitingTimeSeconds(routeParams.minWaitingTimeSeconds),
+    maxTotalTravelTimeSeconds(routeParams.maxTotalTravelTimeSeconds),
+    maxAccessWalkingTravelTimeSeconds(routeParams.maxAccessWalkingTravelTimeSeconds),
+    maxEgressWalkingTravelTimeSeconds(routeParams.maxEgressWalkingTravelTimeSeconds),
+    maxTransferWalkingTravelTimeSeconds(routeParams.maxTransferWalkingTravelTimeSeconds),
+    maxFirstWaitingTimeSeconds(routeParams.maxFirstWaitingTimeSeconds),
+    alternatives(routeParams.alternatives),
+    forwardCalculation(routeParams.forwardCalculation)
+  {
+    scenarioUuid = scenario->uuid;
```
greenscientist: We should copy from the data of the RouteParams and not the passed scenario one

connection_scan_algorithm/src/route_parameters.cpp
greenscientist: Do we need to save the scenario pointer in the object since we extra the data at construction time?

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
...
+    onlyAgenciesIdx = scenario->onlyAgenciesIdx;
+    onlyNodesIdx = scenario->onlyNodesIdx;
+    onlyModesIdx = scenario->onlyModesIdx;
+    exceptLinesIdx = scenario->exceptLinesIdx;
+    exceptAgenciesIdx = scenario->exceptAgenciesIdx;
+    exceptNodesIdx = scenario->exceptNodesIdx;
+    exceptModesIdx = scenario->exceptModesIdx;
+  }
+
+  int getIntegerValue(std::string strValue) {
```
greenscientist: Feel that they should be in an helper class or at least static.

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
...
+      return std::stoi(strValue);
+    }
+    catch (...)
+    {
+      // Throwing an error, but do we want to fallback to default and just ignore this one?
+      throw ParameterException(ParameterException::Type::INVALID_NUMERICAL_DATA);
+    }
+  }
+
+  RouteParameters createRouteODParameter(std::vector<std::pair<std::string, std::string>> &parameters, std::map<boost::uuids::uuid, int> &scenarioIndexesByUuid, std::vector<std::unique_ptr<Scenario>> &scenarios)
```
greenscientist: Maybe a comment of why we should use this function?

greenscientist: Also feel like a factory, should be static so that we don't need to instanciate a RouteParameters to call it

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
...
+    boost::uuids::string_generator uuidGenerator;
+
+    Scenario * scenario = nullptr;
+    boost::optional<boost::uuids::uuid> scenarioUuid = boost::none;
+    boost::uuids::uuid originNodeUuid;
+    boost::uuids::uuid destinationNodeUuid;
+
+    // Initialize default values
+    int timeOfTrip = -1;
+    int minWaitingTimeSeconds = 3 * 60;
```
greenscientist: All those number should be in constant

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,273 @@
...
+    int maxTransferWalkingTravelTimeSeconds = 20 * 60;
+    int maxFirstWaitingTimeSeconds = 30 * 60; // Ignore all connections at access nodes if waiting time would be more than this value.
+    boost::optional<Point> origin = boost::none;
+    boost::optional<Point> destination = boost::none;
+    bool alternatives = false;
+    bool forwardCalculation = true;
+
+    std::vector<std::string> latitudeLongitudeVector;
+
+    for (auto & parameterWithValue : parameters)
```
greenscientist, line 113: Eventually, we should use a parsing library, but fine for now

greenscientist (changes requested):
Lots of small things

connection_scan_algorithm/src/route_parameters.cpp
tahini: I'll add a FIXME but unfortunately yes, because the alternatives calculation tweaks the exceptLine vector and a new object needs to be created. I have a TODO there (in the next yet-unexisting PR) to make this data a calculation parameter, but until then, it's easier to keep the scenario pointer in this object.

connection_scan_algorithm/src/route_parameters.cpp
tahini: definitely, it's just my lack of C++ knowledge that made it this way...

connection_scan_algorithm/src/route_parameters.cpp
tahini: The comment is in the hpp file. Should I declare it as static there too? I guess yes

connection_scan_algorithm/src/route_parameters.cpp
tahini: Can this be a reference? It's a unique_ptr that belongs to someone else, can I pass it around as reference? I'll have to read about this... But I think I tried and failed, but maybe just didn't try hard enough...

tahini:
I'll update this PR once the new route API is accepted.

connection_scan_algorithm/src/route_parameters.cpp
greenscientist: Yes, the static should be in both instance, otherwise I think it's not the same function.

tahini: From my understanding of static functions in C (which is the same in C++?), static is scope limiting, that means if this factory were declared as static, it cannot be called from external file (I read https://stackoverflow.com/questions/42056160/static-functions-declared-in-c-header-files and other info on static keyword). Note that this function is a global function, outside the RouteParameters class. Did you mean it should be a public static method of the RouteParameters class? What would be the advantage versus global function like this?

connection_scan_algorithm/src/route_parameters.cpp
tahini: Another option, instead of saving the scenario pointer would be to add a specific temporary constructor to create the alternatives RouteParameter object. Since it's temporary anyway, I prefer a temporary additional getter than a temporary constructor method (in case there are more such temporary constructors to implement, while the getter is one function and that's that), but it's not a strong opinion.

connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -0,0 +1,280 @@
...
+        destination(std::move(dest_)),
+        scenario(scenario_),
+        timeOfTrip(_timeOfTrip),
+        minWaitingTimeSeconds(minWaitingTime),
+        maxTotalTravelTimeSeconds(maxTotalTime),
+        maxAccessWalkingTravelTimeSeconds(maxAccessTime),
+        maxEgressWalkingTravelTimeSeconds(maxEgressTime),
+        maxTransferWalkingTravelTimeSeconds(maxTransferTime),
+        maxFirstWaitingTimeSeconds(maxFirstWaitingTime),
+        alternatives(alt),
```
greenscientist: Detail, but we could have a more explicit variable name than alternatives. Something that fit more with a boolean, with an action maybe. Something like providesAlternative maybe

connection_scan_algorithm/src/route_parameters.cpp
greenscientist: Static in C is not the same that in C++.   A Static function in a class in C++ means a function that can be called without having to create an instance

tests/connection_scan_algorithm/route_parameters_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+        scenario->uuid = uuidGenerator(EMPTY_SCENARIO_UUID);
+        scenario->name = "Test empty scenario";
+
+        scenarioIndexesByUuid[scenario->uuid] = scenarios.size();
+        scenarios.push_back(std::move(scenario));
+    }
+
+    void TearDown( ) override
+    {
+        // Nothing to do? Shoud we delete the scenarios?
```
greenscientist, line 46: You might want to scenarios.clear() here. Otherwise you keep adding more seach time you call SetUp

tests/connection_scan_algorithm/route_parameters_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+#include "parameters_test.hpp"
+#include "scenario.hpp"
+
+const std::string EMPTY_SCENARIO_UUID = "acdcef12-1111-2222-3333-444455558888";
+
+class RouteParametersFixtureTests : public BaseParametersFixtureTests
+{
+protected:
+    std::vector<std::unique_ptr<TrRouting::Scenario>> scenarios;
+    std::map<boost::uuids::uuid, int> scenarioIndexesByUuid;
```
greenscientist, line 16: WHy not just a map witi UUID:Scenario ?

greenscientist (commented):
Ok, just small comments in the test, otherwise rest looks good.

tests/connection_scan_algorithm/route_parameters_test.cpp
tahini: It's the data type that is used in the trRouting code and that the route parameters factory expects. When we deal with how the data is stored, we can review this. But for now, this is the expected type and the scope of this branch is big enough as it is! (with the patches rewriting the query parameters waiting in line for review)

tests/connection_scan_algorithm/route_parameters_test.cpp
tahini: Done

greenscientist: I would have say that we should remove the comment, but I'm not sure what we need to do with a std::make_unique, if we need to delete it manually of if it's automatic. So it's good to keep a TODO here.

greenscientist (approved):
Good enough. 

Next!

PR #109, opened 2022-02-14
csa: Send the route parameters to the calculation methods

tahini:


connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -97,17 +96,28 @@ namespace TrRouting
...
       {
-        params.maxTotalTravelTimeSeconds = routingResult.travelTimeSeconds + params.alternativesMaxAddedTravelTimeSeconds;
-      }
-      else
-      {
-        params.maxTotalTravelTimeSeconds = maxTravelTime;
+        maxTravelTime = routingResult.travelTimeSeconds + params.alternativesMaxAddedTravelTimeSeconds;
       }
-      
+      // TODO: We should not create a whole new object just to update maxTravelTime. This parameter should be in the calculation specific parameters, which do not exist yet
```
greenscientist, line 105: Or make a copy of the Parameter object and just modify the maxTravelTime

connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -159,9 +159,19 @@ namespace TrRouting
             std::cout << (i+1) << "/" << odTripsCount << std::endl;
           }
         }
-        origin        = odTrip->origin.get();
-        destination   = odTrip->destination.get();
-        routingResult = calculate(true, resetFilters); // reset filters only on first calculation
+        RouteParameters odTripParameters = RouteParameters(std::make_unique<Point>(odTrip->origin.get()->latitude, odTrip->origin.get()->longitude),
+          std::make_unique<Point>(odTrip->destination.get()->latitude, odTrip->destination.get()->longitude),
+          parameters.getScenario(),
```
greenscientist, line 164: Maybe have a "clone" function in RouteParameters which takes a new origin and destination, but keeps everything else and returns a new RouteParameters ?

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -427,17 +422,20 @@ int main(int argc, char** argv) {
           std::cerr << "  -- number of od trips calculated -- " << countOdTripsCalculated << "\n";
         }
       }
+
+    }
+    catch (...)
+    {
+      response = "{\"status\": \"error\", \"error\": \"Wrong or malformed query\"}";
```
greenscientist, line 428: We should really not be generating json "by hand"

tests/connection_scan_algorithm/csa_route_calculation_test.cpp
```diff
@@ -0,0 +1,379 @@
+#include <errno.h>
```
greenscientist, line 1: Shouldn't this file be called csa_v2_route.... ???

greenscientist (approved):
Many small comments. 

Still no happy with the design, I hope we get to something much simplier some day!

connection_scan_algorithm/src/transit_routing_http_server.cpp
tahini: Again, don't be too hasty... it's the next PR on the line...

tests/connection_scan_algorithm/csa_route_calculation_test.cpp
tahini: Not really, even v1 uses this code path now, it's calculation test. v1 should eventually be shrunk to just make sure the parameters array maps to the right RouteParameter object. But since there are still some parameters in the Parameter object, I still want to make sure the complete flow works, so I'm not modifying v1 tests just yet.

connection_scan_algorithm/src/alternatives_routing.cpp
tahini: I contemplated adding copy constructors or clone methods, here and for the od_trip_routing function below. I finally decided not to do it, since they will not be necessary anymore once we have calculation specific parameters, and instead added thos TODOs. I can replace it with clone methods if you prefer.

connection_scan_algorithm/src/transit_routing_http_server.cpp
tahini: Actually, this is legacy, this code is just copy-pasted, I'll update with json object.

connection_scan_algorithm/src/alternatives_routing.cpp
greenscientist: It's fine, if we plan to remove this in the future.

connection_scan_algorithm/src/transit_routing_http_server.cpp
tahini: On second thought, there's a few json string "by hand" in this file, I'll update separately, with the results PR.

tahini:
Closing this PR to reopen it based on #112 to have docker build get triggered from my repo

PR #112, opened 2022-02-17
Replace boost::optional with std::optional

### LINKED ISSUE #111: Does not build with old version of boost
Recent changes added call to optional.has_value, which was only added in boost 1.68. Debian buster have boost 1.67. This function was added to add a compatibility with std::optional (https://www.boost.org/doc/libs/1_74_0/libs/optional/doc/html/boost_optional/relnotes.html)

greenscientist:
We were using functions in boost::optional that were only available in recent version of boost,
breaking builds in older distribution. std::optional is available in CXX17 and provided in gcc since
version 5.

This add build requirement on CXX17 and replace all occurance of boost::optional with the std C++ one.

All unit test still pass

Close #111

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -12,7 +12,7 @@ LT_INIT
 # AX_ macros copied from the autoconf-archive
 
 # TODO Do we need the strict mode or not here ?
-AX_CXX_COMPILE_STDCXX_14(noext)
+AX_CXX_COMPILE_STDCXX(17,noext)
 
 AX_BOOST_BASE
 AX_BOOST_FILESYSTEM
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -23,7 +23,6 @@
 #include <boost/range/adaptor/reversed.hpp>
 //#include <boost/compute.hpp>
 #include <boost/tokenizer.hpp>
-#include <boost/optional.hpp>
 #include <boost/filesystem.hpp>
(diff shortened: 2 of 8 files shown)


tahini:
You seem to have updated googletest stuff also. Is it by design? Otherwise, you can go in the googletest directory and checkout `release-1.11.0`. It was updated in january and I'm not sure how local pulls handle those changes...

greenscientist:
ok, I missed those changes

greenscientist:
Remove the googletest change

tahini:
There's a strange error in dockerhub... Isn't googletest supposed to be public?

greenscientist:
Maybe we just need to autorize external repo somewhere.

greenscientist:
But I think this is related to something to configure in dockerhub. The builds goes ok on my private docker build

tahini:
Can you look at your build if you have anything special configured for googletest?

greenscientist:
I did not see anything

greenscientist:
Currently running the automated build with my branch on my docker account

greenscientist:
Work on my docker repo

tahini:
By the way, why is it even trying to fetch googletest? Shouldn't it be only if we run tests?

tahini:
You can try rebasing on v2c now

greenscientist:
If work! (Make sure to update the v2c in your fork)

tahini (approved):
Thanks! @kaligrafy, do you have any objection to bumping the c++ version from 14 to 17?

PR #113, opened 2022-02-17
csa: Send the route parameters to the calculation methods

tahini:


.gitmodules
```diff
@@ -1,3 +1,3 @@
 [submodule "googletest"]
 	path = googletest
-	url = git@github.com:google/googletest.git
+	url = https://github.com/google/googletest.git
```
tahini: @greenscientist Here I changed the url of the googletest github repo to its https version (no ssh key required I suppose, fully public, i hoped), but dockerhub's build still tries to clone form `git@github.com...` Any idea why?

greenscientist (approved):
If I have approved once, I can do  it a second time

PR #119, opened 2022-02-21
Revert "Update the base Dockerfile to use bullseye"

greenscientist:
This reverts commit dbecfdedae010f4a83bdb906d8e84e61576408ed.


tahini (approved):
ftr, docker build fails with bullseye, there seems to be a conflict in the c++ library. It will be investigated later. Let the progress train continue in the meantime!

PR #120, opened 2022-02-21
csa: Drop csv support in od_trips calculation function

### LINKED ISSUE #110: csa: odTripRouting should return results, and not handle csv
For now, trRouting should drop support of csv for odTrip routing and simply return the results as objects. It is how odTrip routing is handled by the callers anyway.

If support of csv should come back at some point, it can be easily generated from the result objects.

tahini:
Fixes #110

For now, support of csv is dropped entirely and only the json response
is returned by the odTripsRouting method. If csv support is required
again in the future, it can easily be generated from the routing
results.


kaligrafy (approved):

greenscientist (approved):
And if we want to bring it back, we can use a proper csv library instead of doing it manually.

PR #121, opened 2022-02-24
Add new types for routing results

tahini:


tahini:
@greenscientist This PR is not complete, I want to add some unit tests, and the PR that goes above (actually using those types) does not work yet, so there may be a bug here, but I'd like a high level review for now. I'll add the PR that uses it tomorrow.

include/routing_result.hpp
```diff
@@ -3,12 +3,317 @@
...
+        do_accept(visitor);
+        return visitor.getResult();
+    }
+    virtual void do_accept(StepVisitorBase &visitor) const = 0;
+  };
+
+  class StepTransitRoutingResult : public StepRoutingResult {
+  public:
+    boost::uuids::uuid agencyUuid;
+    std::string agencyAcronym;
```
greenscientist, line 67: That's a lot of element here, kind of wonder if they should be regrouped in a subobject

include/routing_result.hpp
```diff
@@ -3,12 +3,317 @@
...
+    virtual void visitNoRoutingFoundResult(const NoRoutingFoundResult& result) = 0;
+  };
+
+  template <class T>
+  class ResultVisitor : public ResultVisitorBase{
+  public:
+    virtual T getResult() = 0;
+  };
+
+  class StepVisitorBase {
```
greenscientist: Not sure of it's role yet, but maybe a Visitor should be in its own file

connection_scan_algorithm/include/result_to_response.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "parameters.hpp"
+
+namespace TrRouting
+{
+  class ResultToResponse {
+  public:
+    virtual nlohmann::json resultToJsonString(RoutingResultNew& result, RouteParameters& params) = 0;
+  };
+
+  class ResultToV1Response: public ResultToResponse {
```
greenscientist: This probably need to be in a separate file

greenscientist (changes requested):
It's missing class description, so we know what are the intents. Probably need to scather more comments around. The visitor class needs to be explained more

A couple of small structural things to adresse

include/routing_result.hpp
```diff
@@ -3,12 +3,317 @@
...
+    std::vector<std::tuple<int, int, int, int, int>> legs; // tuple: tripIdx, lineIdx, pathIdx, start connection index, end connection index
+    SingleCalculationResult():
+      RoutingResultNew(result_type::SINGLE_CALCULATION)
+    {}
+    void do_accept(ResultVisitorBase &visitor) const override {
+      return visitor.visitSingleCalculationResult(*this);
+    }
+  };
+
+  class NoRoutingFoundResult : public RoutingResultNew {
```
tahini: I wonder if we need the NoRouting result, or if it should just be a nullptr...

tahini: Actually a nullptr object cannot be visited, so it's probably better to keep this class

include/routing_result.hpp
tahini: Maybe an optimization for later, but with only the tripUuid and nodeUuid, we can get pretty much all the rest, given that we have access to the data (data, which is now contained within the calculator object, so it is far from a given that access to the data is available where/when we need it). It could be an optimization for later though. Once this is in, we can open a issue to track when/if we want to change this.

tahini:
@greenscientist  I updated as per your comments: more documentation and split visitor/result and converter/converter_to_v1 to separate files. I'll wait for your review of this PR before updating the next one.

include/routing_result.hpp
greenscientist: Do we need to convert a NoRouting result? Why would we need a visitor in this case?

include/routing_result.hpp
```diff
@@ -3,12 +3,323 @@
 
 #include <vector>
 #include <string>
+#include <boost/uuid/uuid.hpp>
+#include <stdlib.h>
+#include "point.hpp"
+#include "routing_result_visitor.hpp"
```
greenscientist, line 8: Could probably just be a forward declaration instead I think

include/routing_result.hpp
```diff
@@ -3,12 +3,323 @@
 
 #include <vector>
 #include <string>
+#include <boost/uuid/uuid.hpp>
+#include <stdlib.h>
+#include "point.hpp"
```
greenscientist: Could probably just be a forward declaration instead I think

include/routing_result.hpp
```diff
@@ -3,12 +3,323 @@
...
+        do_accept(visitor);
+        return visitor.getResult();
+    }
+    virtual void do_accept(ResultVisitorBase &visitor) const = 0;
+  };
+
+  /**
+   * @brief Base class for routing result steps. It accepts a visitor for the steps class hierarchy
+   */
+  class StepRoutingResult {
```
greenscientist: SHould this inherit from RoutingResultNew?

include/routing_result.hpp
```diff
@@ -3,12 +3,323 @@
...
+#include <boost/uuid/uuid.hpp>
+#include <stdlib.h>
+#include "point.hpp"
+#include "routing_result_visitor.hpp"
 
 namespace TrRouting
 {
-  
+  enum result_type { SINGLE_CALCULATION, NO_ROUTING_FOUND, ALTERNATIVES, ALL_NODES };
+  enum result_step_type { WALKING, BOARDING, UNBOARDING };
```
greenscientist, line 17: Shouldn't each of these type their own classes?

greenscientist (commented):
The class hierarchy seems to be lacking the top level. 
Am I missing something? Maybe it's lacking an high level doc/comment?

include/routing_result.hpp
tahini: No, steps are not results. It's really how it's explained.

include/routing_result.hpp
tahini: They sort of are, but since we need to dynamically cast these enums are used to know which class to cast to. Is there a better way?

include/routing_result.hpp
tahini: As explained in the other PR, NoRouting is boring for now, but it will contain a reason later. Unless we use try catch and throw the no routing explanation, but no routing is a valid workflow and using error throwing for it it not necessarily good practice.

Another option would be to include the NoRoutingFound use case in the SingleCalculationResult since it's when it will happen.

What do you think? 3 options:

* the one in this PR (different class)
* try-catch
* merge with SingleCalculation
* other?

include/routing_result.hpp
greenscientist: This ^^

include/routing_result.hpp
greenscientist: and this ^^

include/routing_result.hpp
```diff
@@ -3,12 +3,315 @@
...
+
+  // Walking step types
+  enum walking_step_type { ACCESS, EGRESS, TRANSFER };
+
+  /**
+   * @brief Base class for routing results. It accepts a visitor for the class hierarchy
+   * 
+   * TODO: Will be renamed to RoutingResult in later patch, when the struct is removed and this class is used instead
+   * 
+   * TODO: Even later, when function calls are detricated and split, there will be no use for a class hierarchy. Concrete result types
```
greenscientist, line 27: They could  still be in an hierarchy,  depending on the structure of the result. We'll need to see :)

include/routing_result.hpp
greenscientist: Still wondering about the naming. I don't feel strongly about it, but maybe just RoutingStep ? To differenciate with RoutingResult ?

include/routing_result.hpp
```diff
@@ -3,12 +3,315 @@
...
+    void do_accept(ResultVisitorBase &visitor) const override {
+      return visitor.visitAlternativesResult(*this);
+    }
+  };
+
+  /**
+   * @brief A class describing a single accessible node details, with the time to reach the node
+   * and the number of transfers for the requested departure time
+   */
+  class AccessibleNodesResult {
```
greenscientist: Maybe also not end this in Result if not a RoutingResult ?

connection_scan_algorithm/include/result_to_v1.hpp
```diff
@@ -0,0 +1,22 @@
...
+#include "result_to_response.hpp"
+
+namespace TrRouting
+{
+  /**
+   * @brief Convert a result object to a json object for the version 1 trRouting API, as described in docs/API.yml
+   */
+  class ResultToV1Response: public ResultToResponse {
+  public:
+    nlohmann::json resultToJsonString(RoutingResultNew& result, RouteParameters& params) override;
```
greenscientist: You dropped the virtual here

greenscientist: Also, I would make them static. I don't see a need to instanciate a ResultToV1Response, we can just call the functions

greenscientist (commented):
Still a few small things

connection_scan_algorithm/include/result_to_v1.hpp
tahini: Don't I need to drop the virtual if I'm implementing it? Also, for static calls, you need to know exactly which object to call. What if the code path can support both v1 and v2? sending the converter as base type object allows to just call this method on it and get whatever api version we require.

greenscientist: No, if you drop the virtual, you prohibit further inheritance. The différent is the absence of =0 which makes it virtual pure.

greenscientist: For the static, would need to  analyse the code a little  bit more. Depending on how it's used.

include/routing_result.hpp
tahini: This one yes, can be forward declared

include/routing_result.hpp
tahini: This one no, we are calling methods from these classes.

connection_scan_algorithm/include/result_to_response.hpp
```diff
@@ -0,0 +1,21 @@
+#ifndef TR_RESULT_TO_RESPONSE
+#define TR_RESULT_TO_RESPONSE
+
+#include "json.hpp"
+#include "routing_result.hpp"
+#include "parameters.hpp"
```
tahini: I guess I can forward declare these here too?

greenscientist: Yeah probably. 
The json one is in a diffrent namespace, so might be annoying to forward declare, but the rest looks easy enough

connection_scan_algorithm/include/result_to_v1.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace TrRouting
+{
+
+  /**
+   * @brief Convert a result object to a json object for the version 1 trRouting API, as described in docs/API.yml
+   */
+  class ResultToV1Response: public ResultToResponse {
+  public:
+    virtual nlohmann::json resultToJsonString(RoutingResultNew& result, RouteParameters& params) override;
```
greenscientist: I guess override is a new keyword in C++!

include/routing_result.hpp
greenscientist: grumble. Not great to implement function in a .h, but ok.

connection_scan_algorithm/include/result_to_v1.hpp
greenscientist: For the static
you seem to be using it like this
      ResultToV1Response resultConverter = ResultToV1Response();
            response = resultConverter.resultToJsonString(alternativeResult, routeParams).dump(2);
You are now using any inheritence property here. 
Instead, if it was static, you could simply do
   response = ResultToV1Response::resultToJsonString(alternativeResult, routeParams).dump(2);

greenscientist:
Almost there, a few forward declaration, and the discussion about the static function.

tahini:
I did the static for the result and a bit of forward declaration in result_to_v1.hpp

greenscientist (approved):
Ok, let's go for this!

PR #122, opened 2022-02-25
csa: Update calculation methods to return new result objects

tahini:


tahini:
@greenscientist First 2 commits are part of #121 Only the last one is to be reviewed independently. And it works with Transition!

ftr, I haven't changed any logic anywhere, just the return values of the functions (and how they are built).

It's making use the unique_ptr for the result object and there's a bit of those. I'll [try to] run valgrind on this to make sure it's not introducing any memory leaks. But let me know if there's anything obviously wrong here please!

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -110,19 +110,23 @@ namespace TrRouting
      */
     DataStatus              prepare();
     void                    reset(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
-    RoutingResult           calculate(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
+    // TODO This function supporst both allNodes and simple calculation, which
```
greenscientist: supporst

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -1,23 +1,61 @@
...
-
-    RoutingResult  routingResult;
-    std::string    response;
-    nlohmann::json json;
-    nlohmann::json alternativeJson;
-    
-    if (odTrip != nullptr) // if odTrip is provided
-    {
-      json["odTripUuid"] = boost::uuids::to_string(odTrip->uuid);
+  class LineIdxStepVisitor : public StepVisitor<int> {
```
greenscientist, line 12: What's a LineIdxStepVisitor

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -328,25 +306,20 @@ namespace TrRouting
...
-                stepJson["travelTimeMinutes"]    = Toolbox::convertSecondsToMinutes(transferTime);
-                stepJson["distanceMeters"]       = distance;
-                stepJson["departureTime"]        = Toolbox::convertSecondsToFormattedTime(arrivalTime);
-                stepJson["arrivalTime"]          = Toolbox::convertSecondsToFormattedTime(transferArrivalTime);
-                stepJson["departureTimeSeconds"] = arrivalTime;
-                stepJson["arrivalTimeSeconds"]   = transferArrivalTime;
-                stepJson["readyToBoardAt"]       = Toolbox::convertSecondsToFormattedTime(transferReadyTime);
-                json["steps"].push_back(stepJson);
-                stepsStr += "transfer" + std::to_string(transferTime) + "s" + std::to_string(distance) + "m|";
+                singleResult.get()->steps.push_back(std::make_unique<StepWalkingRoutingResult>(
```
greenscientist: I would add a comment here to explain what it going on

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -5,11 +5,14 @@
...
 
-  RoutingResult Calculator::reverseJourneyStep(RouteParameters &parameters, int bestDepartureTime, int bestAccessNodeIndex, int bestAccessTravelTime, int bestAccessDistance)
+  std::unique_ptr<RoutingResult> Calculator::reverseJourneyStep(RouteParameters &parameters, int bestDepartureTime, int bestAccessNodeIndex, int bestAccessTravelTime, int bestAccessDistance)
   {
-  
-    RoutingResult    result;
-    nlohmann::json   json;
+
+    // FIXME: One code path supports both Single calculation and all nodes, we shouldn't have to deal with both of these types here
+    std::unique_ptr<RoutingResult> result = std::make_unique<NoRoutingFoundResult>();
```
greenscientist: I don't really get why all the unique_ptr here

greenscientist (changes requested):
Please explain the the commit message why you added all the unique_ptr

That plus probably add more comments in general

include/routing_result.hpp
```diff
@@ -3,42 +3,314 @@
...
+    boost::uuids::uuid pathUuid;
+    std::string modeName;
+    std::string mode;
+    boost::uuids::uuid tripUuid;
+    int legSequenceInTrip;
+    int stopSequenceInTrip;
+    boost::uuids::uuid nodeUuid;
+    std::string nodeCode;
+    std::string nodeName;
+    Point& nodeCoordinates;
```
tahini: About the unique_ptr, I'm not sure it still applies, I've been through a lot with this patch... I'll check next week. But the main reason for them is that the copy constructor for these classes didn't work to pass result objects as objects. I don't remember if it was the Point here or the vectors that were causing compilation errors, but the pointers ended up working and it remained this way...

tahini:
So after trying to remove the unique_ptr, it turns out that it is not possible to return by value an object of an abstract type and the return values cannot be of concrete types, as methods can return one of many (for example, no result, alternatives or singleCalculation). The option of reference does not work as the object are created in the function (on the stack) and do not exist once out of the function. Returning a pointer is the way to go, and to avoid memory leaks, unique_ptr are used.
@greenscientist  If you think of any other way, or how to refactor the code to avoid that, let me know!

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -367,131 +345,55 @@ namespace TrRouting
...
+      result = std::move(allNodesResult);
     }
 
     if (params.debugDisplay)
-      std::cerr << "-- forward result: " << result.status << " dts:" << result.departureTimeSeconds << std::endl;
+      std::cerr << "-- forward result: " << result.get()->resType << std::endl;
 
-    result.json = json;
-    return result;
+    return std::move(result);
```
greenscientist: Is this std::move required?

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -110,19 +110,23 @@ namespace TrRouting
...
     std::tuple<int,int,int,int> reverseCalculation(RouteParameters &parameters); // best departure time, best access node index, best access travel time: -1,-1,-1 if non routable, too long or all nodes result
-    RoutingResult           forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, int bestEgressNodeIndex, int bestEgressTravelTime, int bestEgressDistance);
-    RoutingResult           reverseJourneyStep(RouteParameters &parameters, int bestDepartureTime, int bestAccessNodeIndex, int bestAccessTravelTime, int bestAccessDistance);
-    std::string             alternativesRouting(RouteParameters &parameters);
+    std::unique_ptr<RoutingResult> forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, int bestEgressNodeIndex, int bestEgressTravelTime, int bestEgressDistance);
+    std::unique_ptr<RoutingResult> reverseJourneyStep(RouteParameters &parameters, int bestDepartureTime, int bestAccessNodeIndex, int bestAccessTravelTime, int bestAccessDistance);
+    std::unique_ptr<RoutingResult> alternativesRouting(RouteParameters &parameters);
```
greenscientist: Ok, finally found what was bugging me with this. Why all these function return the same type of results? They should be their own result type. This would allow to have a concrete return results, which is a child of RoutingResult. 
Thoughts ?

greenscientist:
@tahini See my latest comment

connection_scan_algorithm/include/calculator.hpp
tahini: I agree, but this would require MAJOR refactoring of the code, and that's outside the scope of this PR or the next or the next next. These methods cover both the allNodes case (for accessibility maps) and the normal calculation.

Why am I doing this PR again? (need to remind myself sometimes...) I want trRouting to have a well-defined API that we can trust (not just look at the result of one query and hope it's always the same) and build upon (and make sure the implementation matches the defined API). 
And then we want to be able to return more specific responses than just no_routing or error. Why is there no_routing? origin and/or destination are not accessible? Are there trips in the scenario around the O/D or just none at all anywhere? So while the NoRoutingFound return type may seem trivial now, it may contain more info later, unless we decide to use error throwing for the various use cases and [eventually later later] have a concrete return type.

tahini:
@greenscientist It is updated with the latest changes to the result classes. As discussed offline, the no routing is now an exception. And the alternative routing returns a concrete type instead of unique_ptr, because it can. So there's only the allNodes/single calculation to eventually split to be able to get only concrete results and no pointers.

connection_scan_algorithm/include/calculator.hpp
greenscientist: Still there

connection_scan_algorithm/include/result_to_response.hpp
```diff
@@ -12,7 +12,7 @@ namespace TrRouting
    */
   class ResultToResponse {
   public:
-    virtual nlohmann::json resultToJsonString(RoutingResultNew& result, RouteParameters& params) = 0;
+    virtual nlohmann::json resultToJsonString(RoutingResult& result, RouteParameters& params) = 0;
```
greenscientist: I don't understand why you drop the New here, speciall that you just added this class

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -4,33 +4,34 @@
...
     }
     else if (bestArrivalTime < MAX_INT) // line found
     {
       foundLine = true;
       resultingNodes.push_back(bestEgressNodeIndex);
+    } else {
+      throw NoRoutingFoundException(NoRoutingFoundException::NO_ROUTING_FOUND);
     }
 
     if (foundLine || params.returnAllNodesResult)
```
greenscientist, line 37: Technically, with the throw you added, this  if is not required anymore. 
(but we should keep a comment stating that)

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -117,21 +116,20 @@ namespace TrRouting
           resultingNodeJourneyStep = forwardJourneysSteps[bestAccessNodeIndex];
           i++;
```
greenscientist, line 117: I don't see this i++ being used. (I know, you did not changed it, but just to reduce line count... :)

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -117,21 +116,20 @@ namespace TrRouting
...
-          json["steps"] = nlohmann::json::array();
           journey.push_back(std::make_tuple(-1,-1,-1,-1,nodesEgressTravelTime[resultingNodeIndex],-1,nodesEgressDistance[resultingNodeIndex]));
         }
         journey.push_front(std::make_tuple(-1,-1,-1,-1,nodesAccessTravelTime[bestAccessNodeIndex],-1,nodesAccessDistance[bestAccessNodeIndex]));
-        
+
         //std::string stepsJson = "  \"steps\":\n  [\n";
-       
+
         i = 0;
```
greenscientist, line 128: The i I mentionned earlier is reset here.

connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -1,14 +1,141 @@
...
+  {
+    nlohmann::json json;
+    // This type of result should not be visited here
+    response = json;
+  }
+
+  void ResultToOdTripJsonVisitor::visitAllNodesResult(const AllNodesResult& result)
+  {
+    nlohmann::json json;
+    // This type of result should not be visited here
```
greenscientist, line 118: Do we need an assert?? The comment make me feel like it

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -395,193 +362,72 @@ namespace TrRouting
...
-      result.firstWaitingTimeSeconds     = -1;
-      result.nonTransitTravelTimeSeconds = -1;
-    }
 
     if (params.returnAllNodesResult)
     {
-      json["numberOfReachableNodes"]  = reachableNodesCount;
-      json["percentOfReachableNodes"] = round(10000 * (float)reachableNodesCount / (float)(nodesCount))/100.0;
+      allNodesResult.get()->numberOfReachableNodes = reachableNodesCount;
+      allNodesResult.get()->percentOfReachableNodes = round(10000 * (float)reachableNodesCount / (float)(nodesCount))/100.0;
```
greenscientist, line 427: Eventuelly, the presence eof 10000 * should be explained here.

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -365,61 +370,72 @@ int main(int argc, char** argv) {
...
+          //file.imbue(std::locale("en_US.UTF8"));
+          file.open(calculator.params.calculationName + ".json", std::ios_base::trunc);
+          file << response;
+          file.close();
+        }
 
-      if (calculator.params.debugDisplay)
-      {
-        for (auto & benchmark : calculator.benchmarking)
+        std::cerr << "-- total -- " << calculator.algorithmCalculationTime.getDurationMicrosecondsNoStop() << " microseconds\n";
```
greenscientist, line 417: Just a "note to shelf", we'll need proper logging at some point!

include/routing_result.hpp
```diff
@@ -24,10 +24,10 @@ namespace TrRouting
    * 
    * TODO: Even later, when function calls are detricated and split, there will be no use for a class hierarchy. Concrete result types
    */
-  class RoutingResultNew {
+  class RoutingResult {
```
greenscientist: I think I would make the renaming from New to not New in a separate commit. Would make this one easier to follow (not necessary if too complicated to extricate)

greenscientist (changes requested):
Mostly nits, comments and naming

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -47,17 +47,15 @@ namespace TrRouting
     return networks.size();
   }
 
-  RoutingResult Calculator::calculate(RouteParameters &parameters, bool resetAccessPaths, bool resetFilters) {
+  std::unique_ptr<RoutingResult> Calculator::calculate(RouteParameters &parameters, bool resetAccessPaths, bool resetFilters) {
```
tahini: Just wondering, could this be a struct union of SingleCalculation, AllNodes and maybe NoRoutingFound instead of unique_ptr and throw NoRoutingFound? Since the alternative routing depends on the noRoutingFound answer, it's like the main workflow depending on try/catch. I see slight performance degradation for alternatives calculations when many there are many alternatives. Could the try/catch be the cause?

greenscientist: Not sure what you are refering to as "this". 
But in general, struct union are a bit hard to understand and more useful to save space than to control flow of a program and more "C" than "C++"

Is the try catch in a loop? Yes, having some error handling is a bit slower. But there's s lot of change in this patch, hard to pinpoint the culprit. Did you generate a flame graph to at least narrow down the function?

tahini: Indeed the try catch is in a loop, to try all possible combinations of lines, so it is expected to have quite a few noRoutingFound results. And I did not do any profiling. For what it's worth, it may be my setup, the difference is not big enough to be worth investigating at this point, but since I wasn't expecting that difference, I doubted...

Let's leave it like this for now, refactor quite a few more things, and then revisit this issue later. Maybe this will be completely different anyway!

connection_scan_algorithm/src/reverse_journey.cpp
tahini: Indeed, why not just *100?

tahini:
@greenscientist  some of your comments are addressed in #123 (renaming the RoutingResultNew class, removing some i's)

tahini:
For the performance issue, I had another hypothesis, which was my own setup (I had enabled kibana and elasticsearch just to test with photon and it surprisingly takes a lot of resources with an empty database and a idle computer...). After disabling them, this has better performances in 45/48 test cases, as expected since we don't create the json response during calculation anymore.

connection_scan_algorithm/src/reverse_journey.cpp
greenscientist: Indeed. hightight why we need a comment :)

connection_scan_algorithm/src/calculator.cpp
greenscientist: Yeah, probably better to run profiling after the end of the refactoring

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -367,131 +346,55 @@ namespace TrRouting
...
-      result.waitingTimeSeconds          = -1;
-      result.accessTravelTimeSeconds     = -1;
-      result.egressTravelTimeSeconds     = -1;
-      result.transferWaitingTimeSeconds  = -1;
-      result.firstWaitingTimeSeconds     = -1;
-      result.nonTransitTravelTimeSeconds = -1;
 
-    }
+    if (params.debugDisplay)
+      std::cerr << "-- forward result: " << (params.returnAllNodesResult ? " allNodes " : "single calculation") << std::endl;
```
greenscientist, line 387: Eventually, this should go throught a proper logging class

greenscientist (approved):
ok, let's go with this. 
if it was new code I would have been more critical, but we are refactoring, so improvment can more later.

PR #123, opened 2022-03-30
csa: Various fixes and updates

tahini:


connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -318,10 +318,10 @@ namespace TrRouting
       {
         std::cout << std::endl;
       }
-      int i {0};
 
       if (params.debugDisplay)
       {
+        int i {0};
```
greenscientist, line 324: Kind of wonder if we need this one to just display en index on the stdout

greenscientist (approved):
Simple enough

connection_scan_algorithm/src/alternatives_routing.cpp
tahini: We'll revisit when we do proper logging...

PR #124, opened 2022-03-30
Remove unnecessary capnp files

tahini:
Doing some cleanup on the capnp cache before migrating the simulations and evolutionary algorithm, and the network and simulation turn out to not be required.


kaligrafy (approved):

kaligrafy:
Some tests are failing though

greenscientist:
Sweet

PR #125, opened 2022-04-01
Use forward declaration in .h instead of includes in calculator.hpp

tahini:
It is a good practice to do so, it avoids including unnecessary headers
in compilation units, it should also decrease compilation times,
especially when make small changes.

Headers are included only in the source files that requires them. This
commit does not significantly reduce compilation time, but the size of
the libcsa.a library has decreased from 107M to 92M.

Also remove a few unnecessary json.hpp includes.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -30,32 +30,7 @@
 #include <limits>
 #include <stdlib.h>
 
-#include "json.hpp"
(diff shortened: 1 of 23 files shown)


greenscientist:
Wow, the size decrease is impressive

greenscientist (approved):
Nothing to say!

tahini:
> Wow, the size decrease is impressive

It is! And I think we can do much better without changing a single line of code. See #126

PR #127, opened 2022-04-01
Add reasons why there is no routing found

tahini:
This adds the reason to the v1 API for the no_routing_found response.

The supported reasons are no access at origin or no access at destination.

### DIFF
--- a/connection_scan_algorithm/src/resets.cpp
+++ b/connection_scan_algorithm/src/resets.cpp
@@ -117,7 +117,7 @@ namespace TrRouting
 
           accessFootpaths = std::move(OsrmFetcher::getAccessibleNodesFootpathsFromPoint(*parameters.getOrigin(), nodes, params.accessMode, params, parameters.getMaxAccessWalkingTravelTimeSeconds(), params.walkingSpeedMetersPerSecond));
           if (accessFootpaths.size() == 0) {
-            throw NoRoutingFoundException(NoRoutingFoundException::NO_ACCESS_AT_ORIGIN); 
+            throw NoRoutingFoundException(NoRoutingReason::NO_ACCESS_AT_ORIGIN); 
           }
         }
       }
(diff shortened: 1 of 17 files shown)


connection_scan_algorithm/src/resets.cpp
```diff
@@ -143,7 +147,7 @@ namespace TrRouting
       }
     }
   
-    if (!params.returnAllNodesResult || arrivalTimeSeconds >= -1)
+    if (!params.returnAllNodesResult || arrivalTimeSeconds > -1)
```
tahini, line 152: @kaligrafy All the tests still pass, but what of this condition? Maybe I changed something, but I'm under the impression that with all_nodes, it would have gotten weird results...

connection_scan_algorithm/src/result_to_v1.cpp
```diff
@@ -199,14 +203,30 @@ namespace TrRouting
     response = json;
   }
 
- nlohmann::json ResultToV1Response::noRoutingFoundResponse(RouteParameters& params)
+ nlohmann::json ResultToV1Response::noRoutingFoundResponse(RouteParameters& params, NoRoutingFoundException& exception)
```
greenscientist: SHould you pass the whole exception or just the reason ? (Feel a bit strange to pass the whole exception

greenscientist (commented):
One commetn about the argument

connection_scan_algorithm/src/result_to_v1.cpp
greenscientist: Main point: If we change the error handling mecanism to not use an exception, your noRoutingFoundResponse would not need to change if you pass the reason directly.

tahini: Indeed, and if we end up encapsulating more information about the reason than just an enum, we can always put it in a class that the exception or other mechanisms can contain.

greenscientist:
Is the commit csa: Increment reachableNodesCount in reverse_journey part if this PR?

greenscientist:
Otherwise the rest looks good. I'll get @kaligrafy answer the question you asked.

tahini:
The relation is that I added a test for accessibility map to make sure it still worked with a valid arrival time. It didn't because there was a bug. It piggy backed on the PR ;-)

greenscientist:
ok. Just checking it was not an branch mix up

greenscientist (approved):
Good from my perspective.

connection_scan_algorithm/src/resets.cpp
kaligrafy: Yeah, good catch! I don't know why there is an equal sign here, but we should test it though

tahini: > Yeah, good catch! I don't know why there is an equal sign here, but we should test it though

I'd like to think that it's now tested while before it wasn't! ;-) What else should be tested? It works with Transition too.

tahini:
@kaligrafy I can merge this?

Issue #126, opened 2022-04-01
housekeeping: Make sure all includes are required


tahini:
For example, in calculator.hpp, there seems to be a lot of system includes and they may not all be necessary. This could reduce the size of the final executable and the compile time to remove unnecessary includes and/or move them to the .cpp file that really requires them.

greenscientist:
@tahini I could probably take an hour looking into that today, if you have not already started

tahini:
I wasn't planning to look at it just yet, but the result of the forward declarations encouraged me to open the issue and leave it for later. Thanks for looking into it!

greenscientist:
A bit more fun to look into C++ code than JS :)

greenscientist:
@tahini Do you think we should look further into this ? There might be a few more classes to look into but the main ones are done.

tahini:
@greenscientist If there's only a few .h to still look at, we may as well keep this open. But if you could just add a comment with which files you did already (or didn't if it's easier), we can come back to this later. Or open an issue with what would be left. Thanks!

greenscientist:
I guess I don't really remember now, it's been a bit too long :/

PR #130, opened 2022-04-02
Cleanup various include

greenscientist:
Did a bunch of clean up in the most frequent use include files. 
(git grep -h "#include" | sort | uniq -c | sort -n)
     24 #include "calculator.hpp"
     33 #include "cache_fetcher.hpp"
     45 #include "parameters.hpp"

This is saving about 20%-25% build time: 
Before: 
real	3m31,590s
user	3m15,584s
sys	0m15,681s

After:
real	2m41,266s
user	2m29,484s
sys	0m11,691s

Also saving a little bit over 2M in the size of the resulting binary. 

Issue #126

### DIFF
--- a/src/nodes_cache_fetcher.cpp
+++ b/src/nodes_cache_fetcher.cpp
@@ -6,15 +6,16 @@
 #include <vector>
 #include <numeric>
 #include <errno.h>
+#include <fcntl.h>
 #include <kj/exception.h>
-
+#include <boost/uuid/string_generator.hpp>
+#include <capnp/serialize-packed.h>
 #include "cache_fetcher.hpp"
 #include "node.hpp"
 #include "point.hpp"
 #include "capnp/nodeCollection.capnp.h"
 #include "capnp/node.capnp.h"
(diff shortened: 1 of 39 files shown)


src/nodes_cache_fetcher.cpp
```diff
@@ -273,4 +274,4 @@ namespace TrRouting
 
 }
 
-#endif // TR_NODES_CACHE_FETCHER
\ No newline at end of file
+#endif // TR_NODES_CACHE_FETCHER
```
tahini, line 277: #ifndef in .cpp files? Is that legit or should we remove them?

tahini (approved):
Very nice! libcsa.a decreased from 92M to 81! And I got it to build in 4:14, while I usually never go below 6min!

src/nodes_cache_fetcher.cpp
greenscientist: hum, it's a bit useless in cpp file true. (But will not cause harm)

PR #132, opened 2022-04-02
More cleanup, this time focused on boost

### LINKED ISSUE #128: housekeeping: Check that all boost lib are used
as a follow up to #126 , we might have many boost libraries that are detected and linked with that are not used.

greenscientist:
We had several boost libraries defined and not used, so let's remove them. 

This build up on PR #130. Remove some unused code that were referencing some boost libs.


tahini:
there's an include of gtfs_fetcher in the benchmark.

greenscientist:
ok, I got confused with all the parallel clean up commit :)

tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -3,13 +3,12 @@
 //Added for the json-example
 #define BOOST_SPIRIT_THREADSAFE
 
+#include <fstream>
 #include <boost/uuid/uuid.hpp>
 #include <boost/uuid/string_generator.hpp>
 
 #include "calculator.hpp"
 #include "cache_fetcher.hpp"
-#include "gtfs_fetcher.hpp" // To initialize calculator object
```
tahini, line 11: These lines should be part of your first commit. If we ever bisect around here, there will be broken commits fwiw...

tahini (approved):
Less dependencies is always good!

tests/benchmark_csa/benchmark_CSA_test.cpp
greenscientist: fixing this...

greenscientist:
Should be ok now

tahini:
Thanks, let's merge!

PR #134, opened 2022-04-04
Fix benchmark after recent changes

tahini:


tests/benchmark_csa/gtest.cpp
```diff
@@ -1,7 +1,20 @@
 #include "gtest/gtest.h" // we will add the path to C preprocessor later
 
+static const char* RUN_ARG = "TR_BENCH";
```
greenscientist: I would do this whole environment check in the Makefile instead of here. 
Having the check in the binary would make it harder to run the benchmark manually.

tahini: > I would do this whole environment check in the Makefile instead of here. Having the check in the binary would make it harder to run the benchmark manually.

What do you mean? The check is done at runtime, so the compilation happens all the time, but if this variable is not set, then the runtime quickly returns nothing, but if set, then it runs the whole benchmark. You can also run manually with `TR_BENCH=1 ./gtest`, but then everything seems outputted to stdout instead of in the gtest.log file... fwiw

greenscientist: hum, I'll need to check again

greenscientist: Here's what I had in mind: 
GIT(boostcleanup) scientist@grrm:~/projets/trRouting$ git diff tests/Makefile.am tests/benchmark_csa/Makefile.am
diff --git a/tests/Makefile.am b/tests/Makefile.am
index c9762315..c0786b46 100644
--- a/tests/Makefile.am
+++ b/tests/Makefile.am
@@ -3,13 +3,8 @@ libgtest_la_SOURCES = ../googletest/googletest/src/gtest-all.cc
 libgtest_la_CPPFLAGS = -I$(top_srcdir)/googletest/googletest/include -I$(top_srcdir)/googletest/googletest
 libgtest_la_LDFLAGS = -pthread
 
-if RUN_BENCHMARK
-MAYBE_BENCHMARK = benchmark_csa
-else
-MAYBE_BENCHMARK =
-endif
-
 SUBDIRS = . \
           cache_fetch \
-          connection_scan_algorithm $(MAYBE_BENCHMARK)
+          connection_scan_algorithm \
+          benchmark_csa
 
diff --git a/tests/benchmark_csa/Makefile.am b/tests/benchmark_csa/Makefile.am
index 9aeebde8..395e229d 100644
--- a/tests/benchmark_csa/Makefile.am
+++ b/tests/benchmark_csa/Makefile.am
@@ -27,6 +27,11 @@ gtest_LDFLAGS = -no-pie -pthread -std=c++17 -lstdc++fs
 
 gtest_CPPFLAGS = -I$(top_srcdir)/googletest/googletest/include -I$(top_srcdir)/googletest/googletest -I$(top_srcdir)/include -I$(top_srcdi>
 
+if RUN_BENCHMARK
 TESTS = gtest
+else
+TESTS =
+endif
+

greenscientist: Here's what I had in mind: 
https://github.com/chairemobilite/trRouting/pull/137

tahini:
Updated with @greenscientist commit.

greenscientist (approved):
Let's go!

PR #137, opened 2022-04-14
Always build benchmark but conditionally run them

greenscientist:
So instead of fully gatting out the build of the benchmark with --enable-benchmark in make check,
run only gate out the running of the tests


tahini (commented):
So we still do ./configure --enable-benchmark, but it's just the runtime that's impacted. But if you want, you can still manually run ./gtest.

That's good. I'll replace my commit with yours in the other PR and update the doc.

greenscientist:
Perfect. I'll abandon this then!

PR #140, opened 2022-05-12
Add accessible nodes and travel times to parameters

tahini:
Those parameters are already used in the code and were just commented
out. They are useful when callers already know the accessible nodes from
the queried locations, to avoid calls to osrm.

The calls to OSRM can take 40% of the time for an accessible map query.


kaligrafy (approved):

kaligrafy:
Good catch!

Issue #141, opened 2022-05-26
Make sure we do not get an infinite loop when creating the journey steps and add tests to validate


kaligrafy:
Make sure we do not create an infinite loop here by verifying that the node index changes at every step
See [forward_journey](https://github.com/chairemobilite/trRouting/blob/6e070306e67d5f0eccf7fb625ab243e40fbcb698/connection_scan_algorithm/src/forward_journey.cpp) line 120 (this should also be checked in reverse journey and everywhere there is a while loop on journey steps).

It created a memory leak when travel time fro some path segments were not integer but were saved in cache as integer.

greenscientist:
Is it possible to create a unit test to reproduce this?

greenscientist:
@kaligrafy @tahini , should we close this now?

tahini:
I think it's not fixed. I think @kaligrafy fixed Transition so the data does not cause this issue, but the issue itself may still happen with some specific data.

kaligrafy:
Let's close for now

kaligrafy:
not fixed yet

PR #142, opened 2022-05-27
verify that data is valid before generating journeys

kaligrafy:
the two while loops in forward and reverse journey can create memory leak (infinite push into a vector) when invalid data is found (like reverse travel times or incorrectly rounded/floored/ceiled travel time seconds that was imported as float, see #141, needs testing though to close issue

### DIFF
--- a/connection_scan_algorithm/src/forward_journey.cpp
+++ b/connection_scan_algorithm/src/forward_journey.cpp
@@ -10,6 +10,7 @@
 #include "agency.hpp"
 #include "routing_result.hpp"
 #include "toolbox.hpp" //MAX_INT
+#include <iostream>
 
 namespace TrRouting
 {
@@ -79,7 +80,7 @@ namespace TrRouting
       int inVehicleDistance        {-1}; int totalInVehicleDistance { 0}; int totalWalkingDistance   { 0};
       int totalTransferDistance    {-1}; int accessDistance         { 0}; int egressDistance         { 0};
       int journeyStepTravelTime    {-1}; int accessWalkingTime      {-1}; int bestAccessNodeIndex    {-1};
-      int transferTime             {-1}; int egressWalkingTime      {-1};
+      int transferTime             {-1}; int egressWalkingTime      {-1}; int bestAccessNodePrevIndex{-1};
       int waitingTime              {-1}; int accessWaitingTime      {-1};
 
       for (auto & resultingNodeIndex : resultingNodes)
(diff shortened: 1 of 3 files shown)


kaligrafy:
need your help for testing!

include/routing_result.hpp
```diff
@@ -321,6 +321,8 @@ namespace TrRouting
     NO_SERVICE_FROM_ORIGIN,
     // There is no service to destination with the query parameters
     NO_SERVICE_TO_DESTINATION
```
tahini: need a comma here

kaligrafy: fixed

tahini:
Why changes in the m4 files? They make the builds fail on linux

kaligrafy:
@greenscientist I could not run reconf or ./configure without auto changing the associated files in the m4 directory. Can you explain what is happening here? I removed the files for now, but I would like to know why they were changed.

kaligrafy:
When I run autoreconf -i on macos, it outputs this and change the two files:
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4'.
glibtoolize: copying file 'm4/libtool.m4'
glibtoolize: copying file 'm4/ltversion.m4'

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -10,6 +10,7 @@
 #include "agency.hpp"
 #include "routing_result.hpp"
 #include "toolbox.hpp" //MAX_INT
+#include <iostream>
```
tahini, line 13: Why this include?

connection_scan_algorithm/src/parameters.cpp
```diff
@@ -19,6 +19,13 @@ namespace TrRouting
     odTripsActivities.clear();
     odTripsModes.clear();
 
+    accessNodesIdx.clear();
```
tahini: I think you merged with the last commit of v2c... Please rebase on v2c to remove this

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -8,6 +8,7 @@
 #include "path.hpp"
 #include "agency.hpp"
 #include "routing_result.hpp"
+#include <iostream>
```
tahini, line 11: Remove?

tahini (changes requested):
I still have bad data lying around, I'll test this on monday.

connection_scan_algorithm/src/reverse_journey.cpp
kaligrafy: cerr needs iostream

connection_scan_algorithm/src/forward_journey.cpp
kaligrafy: needed for cerr

include/routing_result.hpp
```diff
@@ -320,7 +320,9 @@ namespace TrRouting
     // There is no service from origin with the query parameters
     NO_SERVICE_FROM_ORIGIN,
     // There is no service to destination with the query parameters
-    NO_SERVICE_TO_DESTINATION
+    NO_SERVICE_TO_DESTINATION,
+    // Invalid data found, which could create memory leaks or incorrect results
+    INVALID_DATA
```
tahini, line 325: You are adding a new reason for no routing, so you should also add it 

1. in result_to_v1.cpp, in the noRoutingFoundResponse function, it should return an appropriate string message.
2. In the API.yml file, in the reasons for no routing (search for NO_SERVICE_TO_DESTINATION)
3. Is it possible to think of faulty data that could cause this problem to happen? If so, you could add a unit test for this reason (one that would fail with current branch and suceed now). If it is possible to at least describe the kind of data in a dummy unit test in csa_route_calculation_test.cpp, then someone could add a test later

include/routing_result.hpp
```diff
@@ -337,6 +339,20 @@ namespace TrRouting
...
 
+  /**
+   * @brief Exception class thrown when getting invalid data
+   *
+   */
+  class InvalidDataException : public std::exception
+  {
+    public:
+      InvalidDataException(NoRoutingReason reason_) : std::exception(), reason(reason_) {};
+      NoRoutingReason getReason() const { return reason; };
```
tahini, line 350: Nvm my preivous comment depending on the answer here: This is a new exception type, that will be caught by the catch-all clause in transit_routing_http_server and return a "Wrong or malformed query". In that case 

1. It's not really the query who's malformed, it's the data! The returned value will be counterintuitive to the caller
2. It's technically not a NoRoutingReason. You could throw the NoRoutingFoundException with the INVALID_DATA reason, or if you want a new exception, you should put a new enum for those reasons, add it to the API and return a better error message to the user in those cases.

tahini (changes requested):
It works with faulty data. Just make sure it is properly handled, and it's the right type of exception that is thrown (see comments)

greenscientist:
Leaving it here for @kaligrafy for max 6 months and then will close.

greenscientist (changes requested):
Please update

PR #143, opened 2022-06-13
Add a no routing reason for no access at origin and destination

tahini:
It is easily possible to know that there is no access in walking
distance at both origin and destination. So a new exception is thrown
for those cases: NO_ACCESS_AT_ORIGIN_AND_DESTINATION. This gives
additional information to the caller without having to separately make
an extra call with origin and destination reversed to know if there is
access at the other end of the trip.


tahini:
I forgot the return value of the http server...

kaligrafy (approved):

PR #145, opened 2022-06-14
don't include project shortname in custom cache

kaligrafy:
when providing a custom cache directory path, we should not add the project shortname to the path, because it should be included if needed in the custom path, this fix problem with Transition sending custom paths for simulations

### DIFF
--- a/src/cache_fetcher.cpp
+++ b/src/cache_fetcher.cpp
@@ -46,15 +46,17 @@ namespace TrRouting
     if (!params.cacheDirectoryPath.empty()) {
       filePath += params.cacheDirectoryPath + "/";
     }
-    if (!params.projectShortname.empty()) {
+    else if (!params.projectShortname.empty()) {
       filePath += params.projectShortname + "/";
     }
     if (customPath.empty())
     {
+      std::cout << "reading " << (filePath + cacheFilePath) << " cache file" << std::endl;
       return filePath + cacheFilePath;
     }
     else
     {
+      std::cout << "reading " << (filePath + customPath + "/" + cacheFilePath) << " cache file" << std::endl;
       return filePath + customPath + "/" + cacheFilePath;
     }
   }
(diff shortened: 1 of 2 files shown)


src/cache_fetcher.cpp
```diff
@@ -46,15 +46,17 @@ namespace TrRouting
     if (!params.cacheDirectoryPath.empty()) {
       filePath += params.cacheDirectoryPath + "/";
     }
-    if (!params.projectShortname.empty()) {
+    if (params.cacheDirectoryPath.empty() && !params.projectShortname.empty()) {
```
tahini: This should be an else if to the previous if. It makes sure that those 2 ifs are mutually exclusive. You will never enter both. If this is what you mean.

tahini (changes requested):

src/cache_fetcher.cpp
kaligrafy: fixed

kaligrafy: fixed test also

src/cache_fetcher.cpp
```diff
@@ -46,15 +46,17 @@ namespace TrRouting
     if (!params.cacheDirectoryPath.empty()) {
       filePath += params.cacheDirectoryPath + "/";
     }
-    if (!params.projectShortname.empty()) {
+    else if (!params.projectShortname.empty()) {
```
tahini, line 49: Why don't we just send the cacheDirectoryPath, which would include the project shortname in the default case? This looks weird and I always found it hard to figure out where to put the cache when running trRouting manually by default. We should have one cacheDirectoryPath, absolute, document it, and that's it. If there isn't already, we should open an issue. Because this is uselessly complicated imho...

tahini (dismissed):

src/cache_fetcher.cpp
kaligrafy: because json2capnp expect the default to include the project shortname... We should discuss this in an issue

tahini:
Hey you merged even if I removed my approval!

Issue #147, opened 2022-06-15
cannot make check on macos except if removing -lstdc++fs in the makefiles


kaligrafy:
To compile tests on macos, we need to remove -lstdc++fs from every makefile.

kaligrafy:
Error was: `ld: library not found for -lstdc++fs`

kaligrafy:
Here is what I found on [this page](https://git.dynare.org/Dynare/dynare/-/issues/1753):
This hardcodes the library as `-lstdc++fs`, but that library name is specific to GCC. For clang >= 9, there is not specific library to link.

The library name should be autodetected, or set based on the compiler.

kaligrafy:
I still get a deprecated warning though:
`od_trips_cache_fetcher_test.cpp:10:35: warning: 'filesystem' is deprecated: std::experimental::filesystem has now been deprecated in favor of C++17's std::filesystem. Please stop using it and start using std::filesystem. This experimental version will be removed in LLVM 11. You can remove this warning by defining the _LIBCPP_NO_EXPERIMENTAL_DEPRECATION_WARNING_FILESYSTEM macro. [-Wdeprecated-declarations]
namespace fs = std::experimental::filesystem;`

greenscientist:
@kaligrafy Please confirm that the problem is fixed.

PR #152, opened 2022-06-29
Clean up std::filesystem usage

greenscientist:
Use std library instead of the experimental one
and 
clean unncessary link flags. 
#147


tahini (approved):
Sounds good!

PR #158, opened 2022-07-05
Move OSRM related parameters out of the Parameters class

greenscientist:
The OSRM ports and hosts, and also the birdDistanceAccessibilityEnabled field was only used in
the OsrmFetcher static function. They were only initialized at startup.
We move them inside the OsrmFetcher class as static members, which match their usage pattern.

Eventually, it would be better to have an actual OsrmFetcher (or other routing data fetcher) objects,
but that's for a bigger future refactoring

### DIFF
--- a/connection_scan_algorithm/src/resets.cpp
+++ b/connection_scan_algorithm/src/resets.cpp
@@ -114,7 +114,7 @@ namespace TrRouting
         {
           spdlog::debug("  fetching nodes with osrm with mode {}", params.accessMode);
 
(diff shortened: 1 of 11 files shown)


tahini:
Sounds good. Have you tried running with Transition?

greenscientist:
> Sounds good. Have you tried running with Transition?

No. I don't really have a way to test it. I assume the TrRouting are good enough.

tahini:
We've had surprises in the past with things not working as expected with Transition... I'll test before approving

tahini (approved):

greenscientist:
Thanks

PR #159, opened 2022-07-05
Move cacheDirectoryPath parameter inside the CacheFetcher class

greenscientist:
Instead of passing around this information in the Paramaters object, we make it
part of the CacheFetcher object directly. For this, we need to remove the static
qualifier on getFilePath.

Lot of modifications, but the vast majority is to remove the params parameter in
many function call that we don't need anymore

### DIFF
--- a/connection_scan_algorithm/src/transit_routing_http_server.cpp
+++ b/connection_scan_algorithm/src/transit_routing_http_server.cpp
@@ -67,7 +67,6 @@ int main(int argc, char** argv) {
   // setup program options:
   spdlog::info("Starting transit routing on port {} for the data: {}", programOptions.port, programOptions.cachePath);
   
-  algorithmParams.cacheDirectoryPath     = programOptions.cachePath;
   algorithmParams.dataFetcherShortname   = programOptions.dataFetcherShortname;
   OsrmFetcher::osrmWalkingPort        = programOptions.osrmWalkingPort;
   OsrmFetcher::osrmCyclingPort        = programOptions.osrmCyclingPort;
@@ -80,7 +79,7 @@ int main(int argc, char** argv) {
     spdlog::set_level(spdlog::level::debug);
   }
 
-  CacheFetcher cacheFetcher    = CacheFetcher();
+  CacheFetcher cacheFetcher    = CacheFetcher(programOptions.cachePath);
   algorithmParams.cacheFetcher = &cacheFetcher;
 
   Calculator calculator(algorithmParams);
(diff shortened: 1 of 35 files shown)


greenscientist:
I did manage to test it with transition

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -167,67 +166,68 @@ int main(int argc, char** argv) {
...
     int i {0};
     bool correctCacheName {false};
+    //TODO Merge this and the preparations.cpp code
     for(std::string cacheName : cacheNames)
     {
       if (cacheName == "data_sources" || cacheName == "all")
       {
         correctCacheName = true;
-        calculator.updateDataSourcesFromCache(calculator.params, customCacheDirectoryPath);
+        calculator.updateDataSourcesFromCache(customCacheDirectoryPath);
```
tahini, line 175: Do you think we could drop the custom cache directory path too? When you ask to update something, we shouldn't be able to specify just any file, it should be exactly the one that was loaded initially. That would be part of another PR, but I'm asking anyway.

src/cache_fetcher.cpp
```diff
@@ -42,10 +45,10 @@ namespace TrRouting
...
   
-  std::string CacheFetcher::getFilePath(std::string cacheFilePath, Parameters& params, std::string customPath) {
+  std::string CacheFetcher::getFilePath(std::string cacheFilePath, std::string customPath) {
     std::string filePath = "";
-    if (!params.cacheDirectoryPath.empty()) {
-      filePath += params.cacheDirectoryPath + "/";
+    if (!cacheDirectoryPath.empty()) {
+      filePath += cacheDirectoryPath + "/";
     }
     if (customPath.empty())
```
tahini, line 53: Now that we dropped the projectName, there could be a single cache directory in parameter and drop custom stuff also. Unless there's a use case for multiple caches (like simulations the day trRouting is multi-threading, though that may still be a case for multiple instances?)? But then it would be confusing for the request...

src/scenarios_cache_fetcher.cpp
```diff
@@ -47,7 +46,7 @@ namespace TrRouting
 
     spdlog::info("Fetching {} from cache... {}", tStr, customPath);
     
-    std::string cacheFilePath = CacheFetcher::getFilePath(cacheFileName, params, customPath) + ".capnpbin";
+    std::string cacheFilePath = CacheFetcher::getFilePath(cacheFileName, customPath) + ".capnpbin";
```
tahini: Since it is not static anymore, I would have expected this call to be done differently, no?

tahini (commented):
Awesome, I love this! Some thoughts and questions for later, but a comment about the getFilePath call before approving.

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: Yes, would love to do that.  I'm also afraid that someone could do it. But we need to make sure to remove that code part from transition first.

src/scenarios_cache_fetcher.cpp
greenscientist: It does not have to. It means that the function that  will be called is the one from the CacheFetcher class. (Which is itself, so it's a bit redundant.).  If the function was virtual and we'd extend and reimplement in a child class, them  yes, this could cause issue.
Should be easy to mass remove it

src/cache_fetcher.cpp
greenscientist: Basically same answer and the previous comment. This is one I would ask @kaligrafy if he still has a use case for it. (Even if the answer is yes, we could investigate how best implement it)

tahini (approved):

greenscientist:
Created #160 to followup on the customPath

PR #161, opened 2022-07-07
Introduce a base class DataFetcher as parent to CacheFetcher

greenscientist:
Instead of having a if on the data source type in preparation.cpp which calls
different functions, we create the infrastructure to simple parse the type in
the main and instantiate the appropriate Fetcher there and pass it to
the Calculator

This simple some functions and remove need for a few parameters in the Parameters class

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -11,10 +11,11 @@
 #include <boost/uuid/uuid.hpp>
 
 #include "calculation_time.hpp"
+#include "parameters.hpp"
 
 namespace TrRouting
 {
-  class Parameters;
+
   class RouteParameters;
   class Mode;
   class DataSource;
@@ -33,6 +34,7 @@ namespace TrRouting
   class Trip;
   class RoutingResult;
   class AlternativesResult;
+  class DataFetcher;
 
   // tuple representing a connection: departureNodeIndex, arrivalNodeIndex, departureTimeSeconds, arrivalTimeSeconds, tripIndex, canBoard, canUnboard, sequence in trip, lineIndex, blockIndex, canTransferSameLine, minWaitingTimeSeconds (-1 to inherit from parameters)
   using ConnectionTuple = std::tuple<int,int,int,int,int,short,short,int,int,int,short,short>;
@@ -65,7 +67,7 @@ namespace TrRouting
 
     std::map<std::string, int> benchmarking;
 
-    Calculator(Parameters& theParams);
+    Calculator(DataFetcher& dataFetcher);
 
     /**
      * Prepare the data for the calculator and validate that there are at least
@@ -190,8 +192,9 @@ namespace TrRouting
     std::vector<std::vector<std::unique_ptr<float>>> tripConnectionDemands;        // tripIndex: [connectionIndex (sequence in trip): sum of od trips weights using this connection (demand)]
     //std::vector<std::unique_ptr<std::vector<std::unique_ptr<int>>>>   tripIndexesByPathIndex;
 
-    Parameters& params;
+    Parameters params;
     CalculationTime algorithmCalculationTime;
+    DataFetcher &dataFetcher;
 
     OdTrip * odTrip;
 
--- a/connection_scan_algorithm/src/preparations.cpp
+++ b/connection_scan_algorithm/src/preparations.cpp
@@ -1,7 +1,8 @@
 #include "spdlog/spdlog.h"
 #include "calculator.hpp"
 #include "parameters.hpp"
-#include "cache_fetcher.hpp"
+#include "data_fetcher.hpp"
+#include "mode.hpp"
 
 namespace TrRouting
 {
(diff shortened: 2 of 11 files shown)


connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -79,10 +77,16 @@ int main(int argc, char** argv) {
     spdlog::set_level(spdlog::level::debug);
   }
 
-  CacheFetcher cacheFetcher    = CacheFetcher(programOptions.cachePath);
-  algorithmParams.cacheFetcher = &cacheFetcher;
+  DataFetcher *fetcher = 0;
+  if (programOptions.dataFetcherShortname == "cache") {
+    //TODO This is not destroyed on shutdown
```
tahini: What about the fetcher.delete() below?

include/data_fetcher.hpp
```diff
@@ -0,0 +1,264 @@
...
+#include <vector>
+#include <map>
+#include <memory>
+#include <boost/uuid/uuid.hpp>
+
+namespace TrRouting
+{
+    class DataSource;
+    class Household;
+  class Person;
```
tahini, line 14: Some bad alignments in this file, visible in github if not in your editor

include/cache_fetcher.hpp
```diff
@@ -58,7 +61,7 @@ namespace TrRouting
      */
     std::string getFilePath  (         std::string cacheFilePath, std::string customPath = "");
     
-    const std::pair<std::vector<Mode>, std::map<std::string, int>> getModes();
+    virtual const std::pair<std::vector<Mode>, std::map<std::string, int>> getModes();
```
tahini, line 64: Do they really need to be repeated here since they are already in the base class? Maybe remove the docs from the methods since it is a copy of the one from data_fetcher and they may eventually diverge.

include/parameters.hpp
```diff
@@ -11,7 +11,6 @@ namespace TrRouting
 {
 
   class DatabaseFetcher;
```
tahini: Can you remove this too?

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -190,8 +192,9 @@ namespace TrRouting
     std::vector<std::vector<std::unique_ptr<float>>> tripConnectionDemands;        // tripIndex: [connectionIndex (sequence in trip): sum of od trips weights using this connection (demand)]
     //std::vector<std::unique_ptr<std::vector<std::unique_ptr<int>>>>   tripIndexesByPathIndex;
 
-    Parameters& params;
+    Parameters params;
```
tahini, line 195: Is this really needed?

tahini (changes requested):

connection_scan_algorithm/include/calculator.hpp
greenscientist: Yes, calculator need a Parameters object to operate. It was initially given to it in the main().

greenscientist:
(fixing the failing test, I don't know why this make check did not run locally before I submitted the PR)

include/parameters.hpp
greenscientist: Yup!  Someday, we might add DBFetcher :)

include/cache_fetcher.hpp
greenscientist: Yes, the function need to be repeated. The base class has virtual pure functions, here it's an actual implementation. 

About the documentation, yes, it should probably be only in the base class.

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: Forgot to remove the TODO once I did the cleanup :)

tests/connection_scan_algorithm/csa_test_base.hpp
```diff
@@ -4,8 +4,8 @@
 #include <boost/uuid/string_generator.hpp>
 
 #include "gtest/gtest.h"
+#include "dummy_data_fetcher.hpp"
```
tahini, line 7: Did you forget to add this file?

greenscientist: Yes!

connection_scan_algorithm/src/preparations.cpp
```diff
@@ -135,94 +136,80 @@ namespace TrRouting
...
-      // not yet implemented
-    }
-    else if (params.dataFetcherShortname == "csv")
-    {
-      // not yet implemented
+    std::tie(modes, modeIndexesByShortname) = dataFetcher.getModes();
+
+    //updateStationsFromCache();
+    //updateStopsFromCache();
+    ret = updateNodesFromCache();
```
tahini, line 143: These methods will eventually need to be renamed

include/dummy_data_fetcher.hpp
```diff
@@ -0,0 +1,169 @@
...
+  class Station;
+  class Node;
+  class Stop;
+  class Line;
+  class Path;
+  class Scenario;
+  class Trip;
+  class Mode;
+
+  /** Dummy class, to be used in tests
```
tahini: iiuc it doesn't do anything now. But we could (and maybe eventually will have to) use it to fill mock data. If so, maybe add comments here to explain its eventual role in test, or just to say it does nothing at all.

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -79,10 +77,15 @@ int main(int argc, char** argv) {
     spdlog::set_level(spdlog::level::debug);
   }
 
-  CacheFetcher cacheFetcher    = CacheFetcher(programOptions.cachePath);
-  algorithmParams.cacheFetcher = &cacheFetcher;
+  DataFetcher *fetcher = 0;
+  if (programOptions.dataFetcherShortname == "cache") {
```
tahini, line 81: Is this a new program option to specify? Or is it the default?

tahini (approved):
Sounds good, does it require changes to the command line? See the program option comment.

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: It's the default. It was already using this (but elsewhere in the code.

connection_scan_algorithm/src/preparations.cpp
greenscientist: Yeah, it was on my TODO list, but seems like I did not write it down.

include/dummy_data_fetcher.hpp
greenscientist: ok

PR #162, opened 2022-07-07
Remove the save to file feature

greenscientist:
It was possible to ask the route API call to save the result to a file.
It does not really make sense for a remote client to trigger a local action like this.
It's simple for the client to juste save the result to a file. It was transmitted on the socket anyway.

Since we don't need fstream anymore, a little bit of include cleanup was called for


tahini (commented):
No objection here, I'll let @kaligrafy have the final say though...

kaligrafy (approved):
Didn't really use this feature anyway.

greenscientist:
(rebase)

PR #165, opened 2022-07-12
Add const to DataFetcher functions

greenscientist:
Several of the parameters passed to the various function were on fact read only
Add const qualifier to highlight and enforce this fact.

Several accessor had to be changed from [] to at(). With that, since at() validate that the data is
there, several tests need to be changed to actually load the dependent data.

### DIFF
--- a/src/lines_cache_fetcher.cpp
+++ b/src/lines_cache_fetcher.cpp
@@ -20,8 +20,8 @@ namespace TrRouting
   int CacheFetcher::getLines(
     std::vector<std::unique_ptr<Line>>& ts,
     std::map<boost::uuids::uuid, int>& tIndexesByUuid,
-    std::map<boost::uuids::uuid, int>& agencyIndexesByUuid,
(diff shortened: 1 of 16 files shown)


src/lines_cache_fetcher.cpp
```diff
@@ -69,13 +69,13 @@ namespace TrRouting
         std::string agencyUuid {capnpT.getAgencyUuid()};
         
         std::unique_ptr<T> t = std::make_unique<T>();
-
+        spdlog::info("AGENCY UUID {}", agencyUuid);
```
tahini: Do you need the added logging in this file or was it just for debugging purposes?

tahini (approved):
Simple comment on the logging

src/lines_cache_fetcher.cpp
greenscientist: Yeah, that was for debugging, good catch

PR #166, opened 2022-07-15
Remove modeIndexesByShortname by using direct reference to Mode object

greenscientist:
Instead of a vector of mode, we simple use a map of UUID:Mode object.
When a index to a mode was used, we simply store a direct reference to the
refered Mode object.

Since we store a reference, we need to construct some of the objects directly.

(Tested with Transition)

### DIFF
--- a/connection_scan_algorithm/src/resets.cpp
+++ b/connection_scan_algorithm/src/resets.cpp
@@ -6,6 +6,7 @@
 #include "toolbox.hpp" //MAX_INT
 #include "od_trip.hpp"
 #include "routing_result.hpp"
+#include "mode.hpp"
 
 namespace TrRouting
 {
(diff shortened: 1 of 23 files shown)


connection_scan_algorithm/src/resets.cpp
```diff
@@ -274,10 +275,11 @@ namespace TrRouting
         if (tripsEnabled[i] == 1 && parameters.getOnlyNodesIdx()->size() > 0)
         {
           // FIXME: This is not right, it should look for a node, not the mode
-          if (std::find(parameters.getOnlyNodesIdx()->begin(), parameters.getOnlyNodesIdx()->end(), trip->modeIdx) == parameters.getOnlyNodesIdx()->end())
+          // FIXME2: Commented out, since mode is now typed, it won't match
```
tahini, line 278: This is node, not modes?  why comment out? What was this even doing? Probably related to the original fixme... Why note remove entirely then?

tahini (approved):
This looks sound, and now we don't have a vector and index map that could potentially not match

connection_scan_algorithm/src/resets.cpp
greenscientist: Did not remove because of the original FIXME, I guess it should be fixed, but I did not investigate the intend of the original author. 

It's in the filters that are not really used at the moment, so I was not too worried for the moment

Issue #167, opened 2022-07-18
Restore households and places if/when required


greenscientist:
To simplify some of the refactoring work currently done, we have commented out/removed most of the fonctionnalities for Households and Places. This will make less code to manage as we are moving stuff around. 
They are not used anywhere, so no functionalities are lost. We once the code stop moving, it will be easy to add back

greenscientist:
Closing this. I don't forsee a need for them in the future and it was not needed for the last 4 years.

PR #168, opened 2022-07-19
Migrate DataSource vector and index map to a single UUID:DataSource map

greenscientist:
Added == and != operators to DataSource

Since Person and OdTrip now store a reference to a DataSource, we added constructor for those.

While doing this, we observed that Households and Places are not used anywhere in the calculation.
We have commented out those files, references and initialization functions. Those were marked
with a TODO #167 (task number to restore them when needed/refactoring done)

Tested with Transition

### DIFF
--- a/include/parameters.hpp
+++ b/include/parameters.hpp
@@ -6,6 +6,7 @@
 #include <map>
 #include <optional>
 #include <boost/uuid/uuid.hpp>
+#include "data_source.hpp"
 
 namespace TrRouting
 {
@@ -15,6 +16,7 @@ namespace TrRouting
   class Scenario;
   class Node;
   class Mode;
+  class DataSource;
 
   class ParameterException : public std::exception
   {
@@ -149,7 +151,8 @@ namespace TrRouting
       int routingDateYear;   // not implemented, use onlyServicesIdx or exceptServicesIdx for now
       int routingDateMonth;  // not implemented, use onlyServicesIdx or exceptServicesIdx for now
       int routingDateDay;    // not implemented, use onlyServicesIdx or exceptServicesIdx for now
-      int onlyDataSourceIdx;
+      //TODO Make it a reference or not??
+      std::optional<DataSource> onlyDataSource;
       std::vector<int> accessNodesIdx;
       std::vector<int> accessNodeTravelTimesSeconds;
       std::vector<int> accessNodeDistancesMeters;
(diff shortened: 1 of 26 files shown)


include/parameters.hpp
```diff
@@ -149,7 +151,8 @@ namespace TrRouting
       int routingDateYear;   // not implemented, use onlyServicesIdx or exceptServicesIdx for now
       int routingDateMonth;  // not implemented, use onlyServicesIdx or exceptServicesIdx for now
       int routingDateDay;    // not implemented, use onlyServicesIdx or exceptServicesIdx for now
-      int onlyDataSourceIdx;
+    //TODO Make it a reference or not??
+    std::optional<DataSource> onlyDataSource;
```
tahini: check alignments in this file

src/data_sources_cache_fetcher.cpp
```diff
@@ -67,28 +65,28 @@ namespace TrRouting
...
+          case dataSource::DataSource::Type::BIKE_SHARING_TRANSACTIONS: t.type = "bikeSharingTransactions"; break;
+          case dataSource::DataSource::Type::GPS_TRACES               : t.type = "gpsTraces";               break;
+          case dataSource::DataSource::Type::STREET_SEGMENT_SPEEDS    : t.type = "streetSegmentSpeeds";     break;
+          case dataSource::DataSource::Type::UNKNOWN                  : t.type = "unknown";                 break;
         }
 
-        tIndexesByUuid[t->uuid] = ts.size();
-        ts.push_back(std::move(t));
+        //tIndexesByUuid[t.uuid] = ts.size();
+        ts[t.uuid] = t;
```
tahini, line 89: So there is no unique_ptr anymore, so iiuc now we have a map of objects and the `at` method of the map returns a & to the object? This line here copies the entire object? Do we have/need a copy constructor? Has it been tested?

tahini (changes requested):
A c++ question and alignment issues

src/data_sources_cache_fetcher.cpp
greenscientist: Yes, it copies the whole object. 
We don't need an explicit copy constructor, the default one will do the job. 

As for testing, it seems to work with Transition.

include/parameters.hpp
greenscientist: updated

greenscientist:
(and rebased)

tahini (approved):
Looks ok, I'm really really looking forward for multi-threaded trRouting, so by all means, go ahead!!

greenscientist:
> Looks ok, I'm really really looking forward for multi-threaded trRouting, so by all means, go ahead!!

still a long way to go :/

PR #170, opened 2022-07-28
Migrate Agency vector and index map to a single UUID:Agency map

greenscientist:
Added == and != operators to Agency

Line, Trip and Scenario now contain references to an Agency object. This allowed
for simplifications in *_journey files


greenscientist:
Tested with Transition

tahini (approved):
Looks sound. Have you checked the performance of those changes? Agency may not be the one with the most impact and I don't really see anything that can have any even minor hit, but I'm-not-a-compiler! :p

greenscientist:
I ran the benchmark, it does not seem much slower.

PR #173, opened 2022-08-02
Migrate Line vector and index map to a single UUID:Line map

greenscientist:
There was a few cases where the LineIndex was saved but not used, so those where removed.
This include the ConnectionTuple, which included a LineIdx.

In areas where you had a Trip object, the Line object can be accessed directly instead of
going to the full vector

Combination class had to be changed to be able to accommodate std::reference_wrapper.

### DIFF
--- a/connection_scan_algorithm/src/alternatives_routing.cpp
+++ b/connection_scan_algorithm/src/alternatives_routing.cpp
@@ -14,11 +14,11 @@ namespace TrRouting
 {
 
 
-  std::string LineIdxToString(const std::vector<int> & linesList, const std::vector<std::unique_ptr<Line>> &lines) {
+  std::string LinesToString(const std::vector<std::reference_wrapper<const Line>> & linesList) {
     std::stringstream formatter;
 
-    for(auto lineIdx: linesList) {
-      formatter << lines[lineIdx].get()->shortname << " ";
+    for(auto line: linesList) {
+      formatter << line.get().shortname << " ";
     }
     return formatter.str();
   }
@@ -26,27 +26,29 @@ namespace TrRouting
   /**
    * @brief Visitor class to get the line ids used by a step
    * 
-   * TODO See TODO of LineIdxVisitor, when concrete classes are returned, the visitor approach for line id should be re...visited!
+   * TODO See TODO of LineVisitor, when concrete classes are returned, the visitor approach for line id should be re...visited!
    */
-  class LineIdxStepVisitor : public StepVisitor<int> {
+  class LineStepVisitor : public StepVisitor<std::optional<std::reference_wrapper<const Line>>> {
   private:
-    int lineIdx;
-    std::map<boost::uuids::uuid, int>& lineIndexesByUuid;
+    std::optional<std::reference_wrapper<const Line>> stepLine;
+    const std::map<boost::uuids::uuid, Line>& lines;
   public:
-    LineIdxStepVisitor(std::map<boost::uuids::uuid, int>& _lineIndexesByUuid):
-      lineIdx(-1),
-      lineIndexesByUuid(_lineIndexesByUuid) {}
+    LineStepVisitor(const std::map<boost::uuids::uuid, Line>& _lines):
+      lines(_lines) {}
     void visitBoardingStep(const BoardingStep& step) override {
-      lineIdx = lineIndexesByUuid[step.lineUuid];
+      spdlog::debug("Step Visitor line {}", boost::uuids::to_string(step.lineUuid));
+      stepLine = lines.at(step.lineUuid);
     }
     void visitUnboardingStep(const UnboardingStep& step) override {
-      lineIdx = -1;
+      //No line for this step type, don't set result
+      stepLine.reset();
     }
     void visitWalkingStep(const WalkingStep& step) override {
-      lineIdx = -1;
+      //No line for this step type, don't set result
+      stepLine.reset();
     }
-    int getResult() override {
-      return lineIdx;
+    std::optional<std::reference_wrapper<const Line>> getResult() override {
+      return stepLine;
     }
   };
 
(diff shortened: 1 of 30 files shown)


greenscientist:
(Tested with transition)

include/combinations.hpp
```diff
@@ -1,5 +1,6 @@
 // combinations.hpp 
 // from https://stackoverflow.com/a/25497877
+// With later  modifications to adapt for std::reference_wrapper
```
tahini, line 3: Would it be worth adding unit tests for this class? Or are you confident enough with your changes and the stability of this class?

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -154,22 +153,19 @@ namespace TrRouting
...
       lastFoundedAtNum = 1;
 
-      spdlog::debug("fastest line ids: {}", LineIdxToString(foundLinesIdx, lines));
-
+      spdlog::debug("fastest line ids: {}", LinesToString(foundLines));
 
-      combinationsKs.clear();
-      for (int i = 1; i <= foundLinesIdx.size(); i++) { combinationsKs.push_back(i); }
-      for (auto k : combinationsKs)
+      for (int k = 1; k <= foundLines.size(); k++)
```
tahini, line 170: Some comments on what the blocks do in this function would be really nice. It's giving me headaches to understand. It fills the allCombinations vector with all combinations of 1 up to n of the lines found in the best result, so each combination can be added as exceptLines later?

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -179,72 +175,63 @@ namespace TrRouting
...
           try {
             result = calculate(alternativeParameters, false, true);
 
             if (result.get()->resType == result_type::SINGLE_CALCULATION)
             {
               SingleCalculationResult& alternativeCalcResult = dynamic_cast<SingleCalculationResult&>(*result.get());
-              LineIdxVisitor alternativeVisitor = LineIdxVisitor(lineIndexesByUuid);
-              foundLinesIdx = alternativeCalcResult.accept(alternativeVisitor);
-              std::stable_sort(foundLinesIdx.begin(), foundLinesIdx.end());
+              LineIdxVisitor alternativeVisitor = LineIdxVisitor(lines);
```
tahini: Maybe change the name of the Visitor class since it does not return idx anymore?

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -179,72 +175,63 @@ namespace TrRouting
...
+                for (int k = 1; k <= foundLines.size(); k++)
                 {
-                  Combinations<int> combinations(foundLinesIdx, k);
-
+                  Combinations<std::reference_wrapper<const Line>> combinations(foundLines, k);
                   for (auto newCombination : combinations)
                   {
-
-                    newCombination.insert( newCombination.end(), combination.begin(), combination.end() );
+                    // Add current exclusion combination to the new exclusion
```
tahini: Why? what's happening here?

connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -177,7 +176,7 @@ namespace TrRouting
 
     for (auto & line : lines)
     {
-      lineProfiles[line->uuid] = 0.0;
+      lineProfiles[line.first] = 0.0;
```
tahini: .first? Or name it lineIter like below to avoid confusion, as this is not a Line, but a  key, value pair.

include/line.hpp
```diff
@@ -41,8 +41,25 @@ namespace TrRouting
...
+  {
+    return lhs.get() == rhs;
+  }
+
+  inline bool operator==(const std::reference_wrapper<const TrRouting::Line>& lhs, const std::reference_wrapper<const Line>& rhs)
+  {
+    return lhs.get() == rhs.get();
+  }
+  // For sorting and std::map usage
+  inline bool operator<(const std::reference_wrapper<const TrRouting::Line>& lhs, const std::reference_wrapper<const Line>& rhs)
```
tahini, line 59: No need for >, >=, <=?

tests/connection_scan_algorithm/csa_test_base.cpp
```diff
@@ -43,6 +45,8 @@
  * TODO: Mock the bird distance calculations to use completely arbitrary data?
  * */
 void BaseCsaFixtureTests::SetUp() {
+  spdlog::set_level(spdlog::level::debug);
```
tahini: Keep or remove? If keep, add whitespaces to match the rest of the file

tahini (changes requested):
This review has been giving me headaches in the alternatives calculation. Are you confident that all alternative are still returned? Unit tests have very basic tests for those (there's only the except from the first result, not enough lines for more exceptions after that). Maybe in the benchmark, you can output the number of alternatives and make sure they match in both cases?

include/combinations.hpp
greenscientist: The change I did is only doing a copy of the result at the end to remove the optional, so quite that it did not change the previous behavior.

greenscientist:
Yeah, it has been giving me headaches too. :)

As for number of alternative, at least in transition i'm getting the same results as before. 
I did inspect the debug log during the tests and I was getting the same number of alternative. (I enabled debug during the benchmark during my tests and the number matched)

connection_scan_algorithm/src/alternatives_routing.cpp
greenscientist: You are asking about the additional comment or the code change?
(I just changed the code because a std::copy felt easier to understand)

tahini: I'm asking for additional comments on what this does. It's really not obvious what's happening here and since you did the head banging work of understanding the function, you might as well share this in comments ;-)

connection_scan_algorithm/src/alternatives_routing.cpp
greenscientist: Yeah, that's basically it

include/line.hpp
greenscientist: No,  sort and std::map only rely on <

tests/connection_scan_algorithm/csa_test_base.cpp
greenscientist: I kind of like having full debug in the tests output. So if you don't might I'll leave it there (propelly aligned)

include/combinations.hpp
greenscientist: Adding basic tests

connection_scan_algorithm/src/alternatives_routing.cpp
greenscientist: Tried adding a few comments, but I think the code could be simplified eventually.

greenscientist:
Some fix.
There was already some test checking the number of alternative result, so that should at least do some basic coverage. (In addition to the manual comparison I have done in the benchmark)

tests/connection_scan_algorithm/combinations_test.cpp
```diff
@@ -0,0 +1,72 @@
+#include "gtest/gtest.h"
+#include "combinations.hpp"
+
+/*class CombinationTests : public ::testing::Test
```
tahini: Remove commented code

connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -175,10 +174,12 @@ namespace TrRouting
...
     json["odTrips"] = nlohmann::json::array();
 
-    for (auto & line : lines)
+    // Initialize lineProfiles
+    for (auto & linePair : lines)
     {
-      lineProfiles[line->uuid] = 0.0;
+      lineProfiles[linePair.first] = 0.0;
     }
+    //TODO Why is it 28 ? That seems too many hours in a day
```
tahini, line 182: In transit, it is habitual to have schedules from midnight to 4AM the next day, whence 28. But I don't know how common it is, if it is a rule, if it works everywhere, if it should rather be configured per agency, etc. @kaligrafy may know

tahini (approved):
Some remaining comments that you may fix before merging or not.

connection_scan_algorithm/src/od_trips_routing.cpp
greenscientist: ok. make sense, but having it hardcoded seems wrong. As you say, we should either infer from the available data. 
I imagine not all agency will have that.

PR #174, opened 2022-08-03
    Migrate Path vector and index map to a single UUID:Path map

greenscientist:


greenscientist:
I'll rework the Line commit, but if you want to take some advance and check the Path already, you can go ahead. It's much simpler. 

Tested with Transition, test and benchmarks

connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -313,33 +311,36 @@ namespace TrRouting
             for (auto & leg : visitor.getLegs())
             {
               legTripIdx            = std::get<0>(leg);
-              const Line &legLine   = std::get<1>(leg);
-              legPathIdx            = std::get<2>(leg);
-              legPath               = paths[legPathIdx].get();
+              //std::reference_wrapper<const Line> lineref =  std::get<1>(leg);
```
tahini, line 314: Remove commented code in this block

include/cache_fetcher.hpp
```diff
@@ -106,8 +106,7 @@ namespace TrRouting
     );
 
     virtual int getPaths(
-      std::vector<std::unique_ptr<Path>>& ts,
-      std::map<boost::uuids::uuid, int>& tIndexesById,
+                         std::map<boost::uuids::uuid, Path>& ts,
```
tahini: Align with next lines

include/routing_result.hpp
```diff
@@ -9,6 +9,7 @@
 #include "json.hpp"
 #include "point.hpp" //Not using a forward declaration, as we use it more directly, see issue #129
 #include "line.hpp" //For Leg
+#include "path.hpp" //For Leg
```
tahini: Here do you need the include or can you forward declare the class since it's a reference_wrapper?

tests/cache_fetch/schedules_cache_fetcher_test.cpp
```diff
@@ -16,11 +16,10 @@ class ScheduleCacheFetcherFixtureTests : public BaseCacheFetcherFixtureTests
 protected:
     std::vector<std::unique_ptr<TrRouting::Trip>> trips;
     std::map<boost::uuids::uuid, TrRouting::Line> lines;
-    std::vector<std::unique_ptr<TrRouting::Path>> paths;
+     std::map<boost::uuids::uuid, TrRouting::Path> paths;
```
tahini: alignment

tahini (commented):
I only reviewed the last commit, looks good, a few minor comments, but should be good after the Line vector PR has been merged and this one rebased

include/routing_result.hpp
greenscientist: apparently I can forward declare. I thought I would need to include it everywhere else.

greenscientist:
rebase (and fixed the details)

tahini (approved):

kaligrafy:
Please make sure you test the performance, since when I changed all hashmaps to vectors, it was wayyyy faster!

kaligrafy:
I heard some times ago that you can tell the compiler to use vectors with indexes internally for some specific hashmaps to increase performance but keep the code clearer. I have absolutely no idea how to implement ths though.

greenscientist:
The performance difference is not significant so far. Some times  a few percentage slower. But the current method is too error prone to be sustainable, so it's ok to take a performance it. 
We can then investigate performance later. 
Also now, in many case we have a direct reference to the data, we don't do a map lookup to access it.

tahini:
Still approved.

greenscientist:
(fyi, I just did a fake update to trigger the codeql action that I
reactivated)

PR #178, opened 2022-09-08
Update and fix APIv2 documentation

tahini:
Before adding new paths to the API, a cleanup of its current state is necessary.

Parameters and routeResponses are moved to their own files for better encapsulation and parameter re-use.

Also, there was a misunderstanding of OpenAPI's polymorphism and the response types were not quite correct. This is fixed here. Swagger now shows the correct documentation.


greenscientist (commented):
Seems ok. 

About the split into version directory, this would mean we don't share anything between v1 and v2 ?

Also, should we rename API.yml to APIv1.yml?

tahini:
I could rename APIv2.yml to API.yml otherwise it's redundant with the directory. Also indeed, we should not share anything between the 2, even if they do right now have some shared fields. But it's an API, we don't want to "accidently" break v1 if we change the still unstable v2.

tahini:
I downgraded the package versions to react-markdown 6.0.3 and remark-gfm to 1.0.0. It appears to work, at least the tildes are displayed in the modal and the tests pass! Unless there's a reason for the latest packages, we can just squash those 2 commits and update commit messages. And open an issue to support esm in jest (it will be necessary eventually for sure...)

greenscientist:
yeah good points

docs/APIv2/API.yml
```diff
@@ -0,0 +1,67 @@
...
+info:
+  title: TrRouting
+  version: '2.0-beta'
+servers:
+  - url: http://localhost:4000
+
+paths:
+  /versions:
+    get:
+      description: Get the versions of the API supported by this inVehicleDistanceMeters
```
greenscientist: It was there already, but this "inVehicleDistanceMeters" looks weird

greenscientist (approved):
See my small comment. can be fixed later

docs/APIv2/API.yml
tahini: Wow, this looks totally weird!

greenscientist: sounds better now

Issue #177, opened 2022-09-07
Add an API call to get all lines that can be taken between origin and destination


kaligrafy:
For ARTM 2022: add an API call to get all alternatives and return array of all possible lines (uuid, shortname, longname, agency uuid, agency acronym and agency name)

Will be called from evolution, to validate bus, subway and train lines declared by respondents.

tahini:
This API call would receive parameters similar to the route endpoint, but instead of returning the complete detailed alternatives list for the trip, it will return only an array of line objects.

@kaligrafy Does it need additional data? Like the number of alternatives in which this line appears? The total number of alternatives found?

kaligrafy:
If it is easy, it would be great to get the number of alternatives matches and the total number of alternatives, but not essential for evolution right now.

greenscientist:
Please provide a complete documentation for the API first.

tahini:
of course ;-) It should be documented in the APIv2.yml file.

greenscientist:
Also, don't implement right away please, code is too much in flux at the moment.

tahini:
I was thinking of naively using the alternatives calculation and squash the response from it, so it won't interfere with anything you're doing, unless you mean something else by "in flux"?

greenscientist:
Current diffstat: 
	modifié :         connection_scan_algorithm/include/calculator.hpp
	modifié :         connection_scan_algorithm/src/Makefile.am
	modifié :         connection_scan_algorithm/src/calculator.cpp
	modifié :         connection_scan_algorithm/src/forward_calculation.cpp
	modifié :         connection_scan_algorithm/src/forward_journey.cpp
	modifié :         connection_scan_algorithm/src/initializations.cpp
	modifié :         connection_scan_algorithm/src/od_trips_routing.cpp
	modifié :         connection_scan_algorithm/src/optimize_journey.cpp
	modifié :         connection_scan_algorithm/src/parameters.cpp
	modifié :         connection_scan_algorithm/src/preparations.cpp
	modifié :         connection_scan_algorithm/src/resets.cpp
	modifié :         connection_scan_algorithm/src/reverse_calculation.cpp
	modifié :         connection_scan_algorithm/src/reverse_journey.cpp
	modifié :         connection_scan_algorithm/src/transit_routing_http_server.cpp
	modifié :         googletest (contenu non suivi)
	modifié :         include/cache_fetcher.hpp
	modifié :         include/data_fetcher.hpp
[...]

PR #179, opened 2022-09-19
Add an API endpoint documentation to get line summary

tahini:
Documents #177

* Add a API endpoint called `v2/summary`, taking the same parameters as the `route` endpoint, along with an object to state what to summarize (only line will be implemented for now). This will return an array of lines and their agency data, for all lines that were returned by the route calculation, as well as the number of alternatives they were part of.
* Move some common responses to the commonResponse.yml file


kaligrafy (approved):

docs/APIv2/summaryResponse.yml
```diff
@@ -0,0 +1,68 @@
...
+  properties:
+    origin:
+      type: string
+      pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+      description: Comma-separated longitude/latitude coordinates of the origin, in the WSG84 coordinates system
+    destination:
+      type: string
+      pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
+      description: Comma-separated longitude/latitude coordinates of the destination, in the WSG84 coordinates system
+    requestTime:
```
greenscientist: Check name of field in query,  might be better

docs/APIv2/routeResponse.yml
```diff
@@ -70,14 +70,20 @@ baseRouteResponse:
   type: object
   properties:
     origin:
-      type: string
-      pattern: '^\-?\d{1,3}(\.\d+)?,\-?\d{1,3}(\.\d+)?$'
-      description: Comma-separated longitude/latitude coordinates of the origin, in the WSG84 coordinates system
+      type: array
```
greenscientist, line 32: Curious, is it possible to "typedef" a lat/long type so we can reuse it in many places?

greenscientist (approved):

docs/APIv2/routeResponse.yml
tahini: I'm not sure, we can typedef an object, that's where the #ref is used, but we can't typedef a type. At least I haven't found it

PR #181, opened 2022-09-20
Implement the `v2/route` and `v2/summary` endpoints

### LINKED ISSUE #177: Add an API call to get all lines that can be taken between origin and destination

tahini:
Fixes #177 

@greenscientist I don't think it has any impact on the refactoring you are doing...

### DIFF
--- a/connection_scan_algorithm/src/transit_routing_http_server.cpp
+++ b/connection_scan_algorithm/src/transit_routing_http_server.cpp
@@ -24,6 +24,7 @@
 #include "program_options.hpp"
 #include "result_to_v1.hpp"
 #include "result_to_v2.hpp"
+#include "result_to_v2_summary.hpp"
 #include "routing_result.hpp"
 #include "osrm_fetcher.hpp"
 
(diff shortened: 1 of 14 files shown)


greenscientist:
Yeah,should not conflict much.
(I'll check the code  in more depth tomorrow)

kaligrafy:
I'll let @greenscientist review this one

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -55,6 +56,45 @@ std::string intializeResponse(Calculator::DataStatus status)
   }
 }
 
+std::string intializeResponseNew(Calculator::DataStatus status)
```
greenscientist: I don't see how the function name match what it does. Seems to convert a Status object to json

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -404,6 +444,70 @@ int main(int argc, char** argv) {
 
   };
 
+  // Routing request for a single origin destination
+  server.resource["^/v2/route[/]?$"]["GET"]=[&server, &calculator, &dataStatus](std::shared_ptr<HttpServer::Response> serverResponse, std::shared_ptr<HttpServer::Request> request) {
+
+    std::string response = intializeResponseNew(dataStatus);
```
greenscientist: Is it possible to put most of this code in a function? Would be easier to move around instead of putting it directly into the parsing code.  It fees like there's a lot of copy/paste here. Would at least add comments explaining the copy/paste.

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -508,6 +509,71 @@ int main(int argc, char** argv) {
 
   };
 
+  // Request a summary of lines data for a route
+  // TODO Copy pasted from v2/route. There's a lot in common, it should be extracted to common class, just the response parser is different
+  server.resource["^/v2/summary[/]?$"]["GET"]=[&server, &calculator, &dataStatus](std::shared_ptr<HttpServer::Response> serverResponse, std::shared_ptr<HttpServer::Request> request) {
```
greenscientist, line 515: Basically same comment about puting in a function. At least you have a comment here.

greenscientist (changes requested):
Two things to revisit, I'll let you comment if it's possible to do or not.

connection_scan_algorithm/src/transit_routing_http_server.cpp
tahini: Good point, that was taken from a branch I started months ago... I'll update

connection_scan_algorithm/src/transit_routing_http_server.cpp
tahini: Possible yes, desirable, yes too. But whatever part of the code is moved out should be web server independent, so we'd need to determine where to put the split, how much can be re-used from v1 to v2 (technically, the difference is only the API, so much should be common). From my previous experimentations with this function, I think it could be easier when the routingResult doesn't have to be a unique_ptr (probably after all the refactoring is done). For now, I think I'll just add a comment... and hopefully get to rewrite this before there,s too much new APIs.

greenscientist (approved):

PR #182, opened 2022-09-29
Initialize global parameter object at startup

tahini:
Initializing those parameters is required for proper calculation as it still contains a few fields that are not really parameters, but are there (like accessMode and egressMode for footpath calculations).

This fixes instances planned to run only V2 queries, as this param object is not re-initialized by those requests, only the `transit/v1/route` query did it. v2 endpoints all returned no routing unless a v1 query was done first.


greenscientist:
Since it's done here, should it be removed from the V1 query?

tahini:
I am not sure, so I would say no... the accessible node query may require this to be re-initialized for each query, while this is not yet supported in v2.

greenscientist (approved):
ok, ugly, but I'll accept

PR #184, opened 2022-10-20
Migrate Node vector and index to a UUID:Node map

greenscientist:
Currently benchmarks are about 2x slower than before. Main time is spent doing the looking into the map<uuid:int> for some query specificy computation. It was speed up significantly by using unordered_map instead of map for those. We also added a "local" only "int" id to replace the uuid lookup. (Maybe we won't need this extra id with further optimisations.)

There's probably more optimisations to be done, but I think we can wait until we restructure the code to be able to run multiple queries in parallel.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -4,6 +4,7 @@
 #include <string>
 #include <vector>
 #include <map>
+#include <unordered_map>
 #include <memory>
 #include <deque>
 #include <tuple>
@@ -12,6 +13,8 @@
 
 #include "calculation_time.hpp"
 #include "parameters.hpp"
+#include "connection.hpp"
+#include "node.hpp"
 
 namespace TrRouting
 {
@@ -26,7 +29,6 @@ namespace TrRouting
   class Agency;
   class Service;
   class Station;
-  class Node;
   class Line;
   class Path;
   class Scenario;
@@ -35,8 +37,7 @@ namespace TrRouting
   class AlternativesResult;
   class DataFetcher;
 
-  // tuple representing a connection: departureNodeIndex, arrivalNodeIndex, departureTimeSeconds, arrivalTimeSeconds, tripIndex, canBoard, canUnboard, sequence in trip, canTransferSameLine, minWaitingTimeSeconds (-1 to inherit from parameters)
-  using ConnectionTuple = std::tuple<int,int,int,int,int,short,short,int,short,short>;
+  using JourneyStep = std::tuple<int,int,int,int,short,int>; //final enter connection, final exit connection, final trip index, transfer travel time, is same node transfer (first, second, third and fourth values = -1 for access and egress journeys)
 
   class Calculator {
 
@@ -82,16 +83,16 @@ namespace TrRouting
     // a concrete result object instead of pointer (that alternatives could use directly), but still
     // use common calculation functions
     std::unique_ptr<RoutingResult> calculate(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
-    std::tuple<int,int,int,int> forwardCalculation(RouteParameters &parameters); // best arrival time,   best egress node index, best egress travel time: MAX_INT,-1,-1 if non routable, too long or all nodes result
-    std::tuple<int,int,int,int> reverseCalculation(RouteParameters &parameters); // best departure time, best access node index, best access travel time: -1,-1,-1 if non routable, too long or all nodes result
+    std::optional<std::tuple<int, std::reference_wrapper<const Node>>> forwardCalculation(RouteParameters &parameters); // best arrival time,   best egress node
+    std::optional<std::tuple<int, std::reference_wrapper<const Node>>> reverseCalculation(RouteParameters &parameters); // best departure time, best access node
     // TODO See calculate
-    std::unique_ptr<RoutingResult> forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, int bestEgressNodeIndex, int bestEgressTravelTime, int bestEgressDistance);
+    std::unique_ptr<RoutingResult> forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, std::optional<std::reference_wrapper<const Node>> bestEgressNode);
     // TODO See calculate
-    std::unique_ptr<RoutingResult> reverseJourneyStep(RouteParameters &parameters, int bestDepartureTime, int bestAccessNodeIndex, int bestAccessTravelTime, int bestAccessDistance);
+    std::unique_ptr<RoutingResult> reverseJourneyStep(RouteParameters &parameters, int bestDepartureTime, std::optional<std::reference_wrapper<const Node>> bestAccessNode);
     AlternativesResult alternativesRouting(RouteParameters &parameters);
     std::string             odTripsRouting(RouteParameters &parameters);
 
-    std::vector<int>        optimizeJourney(std::deque<std::tuple<int,int,int,int,int,short,int>> &journey);
+    std::vector<int>        optimizeJourney(std::deque<JourneyStep> &journey);
 
     /**
      * The update*FromCache methods get the data from the cache
@@ -154,8 +155,8 @@ namespace TrRouting
     std::map<boost::uuids::uuid, Service>    services;
     const std::map<boost::uuids::uuid, Service> & getServices() {return services;}
 
-    std::vector<std::unique_ptr<Node>>       nodes;
-    std::map<boost::uuids::uuid, int>        nodeIndexesByUuid;
+    std::map<boost::uuids::uuid, Node>       nodes;
+    const std::map<boost::uuids::uuid, Node> & getNodes() {return nodes;}
 
     std::map<boost::uuids::uuid, Line>       lines;
     const std::map<boost::uuids::uuid, Line> & getLines() {return lines;}
@@ -181,8 +182,7 @@ namespace TrRouting
 
   private:
 
-    enum connectionIndexes : short { NODE_DEP = 0, NODE_ARR = 1, TIME_DEP = 2, TIME_ARR = 3, TRIP = 4, CAN_BOARD = 5, CAN_UNBOARD = 6, SEQUENCE = 7, CAN_TRANSFER_SAME_LINE = 8, MIN_WAITING_TIME_SECONDS = 9 };
-    enum journeyStepIndexes: short { FINAL_ENTER_CONNECTION = 0, FINAL_EXIT_CONNECTION = 1, FINAL_TRANSFERRING_NODE = 2, FINAL_TRIP = 3, TRANSFER_TRAVEL_TIME = 4, IS_SAME_NODE_TRANSFER = 5, TRANSFER_DISTANCE = 6 };
+    enum journeyStepIndexes: short { FINAL_ENTER_CONNECTION = 0, FINAL_EXIT_CONNECTION = 1, FINAL_TRIP = 2, TRANSFER_TRAVEL_TIME = 3, IS_SAME_NODE_TRANSFER = 4, TRANSFER_DISTANCE = 5 };
 
     int              departureTimeSeconds;
     int              initialDepartureTimeSeconds;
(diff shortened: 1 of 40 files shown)


greenscientist:
(Don't  fully review yet, there's tons of commented out code around.)

greenscientist:
Also need to figure out why the mac os compiler is unhappy

greenscientist:
Last flamegraph to have an idea of the bottlenecks
profile7

greenscientist:
Fun...  The build issue might be a bug in clang: https://github.com/llvm/llvm-project/issues/48582

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -82,16 +85,16 @@ namespace TrRouting
...
-    std::tuple<int,int,int,int> forwardCalculation(RouteParameters &parameters); // best arrival time,   best egress node index, best egress travel time: MAX_INT,-1,-1 if non routable, too long or all nodes result
-    std::tuple<int,int,int,int> reverseCalculation(RouteParameters &parameters); // best departure time, best access node index, best access travel time: -1,-1,-1 if non routable, too long or all nodes result
+    std::optional<std::tuple<int, std::reference_wrapper<const Node>>> forwardCalculation(RouteParameters &parameters); // best arrival time,   best egress node
+    std::optional<std::tuple<int, std::reference_wrapper<const Node>>> reverseCalculation(RouteParameters &parameters); // best departure time, best access node
     // TODO See calculate
-    std::unique_ptr<RoutingResult> forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, int bestEgressNodeIndex, int bestEgressTravelTime, int bestEgressDistance);
+    std::unique_ptr<RoutingResult> forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, std::optional<std::reference_wrapper<const Node>> bestEgressNode);
```
tahini, line 89: so the single reference to the node replaces the best node index, travel time and distance?

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -27,27 +27,34 @@ namespace TrRouting
...
     bool             foundLine            {false};
 
-    std::vector<int> resultingNodes;
+    std::vector<std::reference_wrapper<const Node>> resultingNodes;
     if (params.returnAllNodesResult)
     {
-      nodesCount     = nodes.size();
-      resultingNodes = std::vector<int>(nodesCount);
-      std::iota (std::begin(resultingNodes), std::end(resultingNodes), 0); // generate sequencial indexes of each nodes
+      ////TODO Do we copy all nodes???
```
tahini: Why would we?

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -27,27 +27,34 @@ namespace TrRouting
...
+      spdlog::debug("!!!! forward journey - returnAllNodesResult not implemented");
+
+      //TODO DOIT maybe?!
```
tahini: Here too, please explain why you add this todo, so we all understand it later

include/path.hpp
```diff
@@ -16,24 +17,44 @@ namespace TrRouting
...
       line(aline),
       direction(adirection),
       internalId(ainternalId),
-      nodesIdx(anodesIdx),
+      nodesRef(anodesRef),
       tripsIdx(atripsIdx),
       segmentsTravelTimeSeconds(asegmentsTravelTimeSeconds),
       segmentsDistanceMeters(asegmentsDistanceMeters) {}
+
+    Path(const boost::uuids::uuid &auuid,
```
tahini, line 35: Why a new constructor? Maybe add documentation for in which case to use each.

tahini (commented):
I did a quick scroll of the PR, nothing seems out of order. I'll do a more "thorough" review after the cleanup is finished.

connection_scan_algorithm/include/calculator.hpp
greenscientist: The travel time and distance were not used anywhere

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: We need them in resultingNodes, when we compute result for all nodes.

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: This was temporary comments to remind me to implement it at the end. It was done.

include/path.hpp
greenscientist: will do. (Basically a transition measure)

greenscientist:
cleaned up...

greenscientist:
tested the time difference between the unordered_map<boost::uuid,int> vs unordered_map<int, int> and that gives us a 2x speed up. So it's worth the small additionnal complexity.

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -80,21 +82,25 @@ namespace TrRouting
...
         {
-          spdlog::debug("bestArrivalTime after forward journey: ", bestArrivalTime);
+          spdlog::debug("bestArrivalTime after forward journey: {}", bestArrivalTime);
           
           arrivalTimeSeconds = bestArrivalTime;
           
           for (auto & egressFootpath : egressFootpaths) // reset nodes reverse tentative times with new arrival time:
           {
-            nodesReverseTentativeTime[std::get<0>(egressFootpath)] = arrivalTimeSeconds - std::get<1>(egressFootpath);
+            nodesReverseTentativeTime[egressFootpath.node.uid] = arrivalTimeSeconds - egressFootpath.time;
```
tahini, line 91: You are not using the uid_t here?

greenscientist: hum, yes, uid is of type uid_t

tahini: oh right... uid not uuid, sorry... id, uid, uuid... next is uuuid? :p

greenscientist: Yeah, I'm not super happy with the uid name, but could not imagine something better then.

connection_scan_algorithm/src/initializations.cpp
```diff
@@ -43,35 +43,13 @@ namespace TrRouting
 
   void Calculator::initializeCalculationData() {
     nodesTentativeTime.clear();
-    nodesTentativeTime.shrink_to_fit();
```
tahini, line 46: These are all local variables where necessary now?

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -119,11 +119,12 @@ namespace TrRouting
             if (std::get<connectionIndexes::CAN_BOARD>(**connection) == 1 && tripsExitConnection[tripIndex] != -1)
             {
               // get footpaths for the arrival node to get transferable nodes:
-              nodeDepartureIndex              = std::get<connectionIndexes::NODE_DEP>(**connection);
+              const Node &nodeDeparture = std::get<connectionIndexes::NODE_DEP>(**connection);
               connectionDepartureTime         = std::get<connectionIndexes::TIME_DEP>(**connection);
               connectionMinWaitingTimeSeconds = std::get<connectionIndexes::MIN_WAITING_TIME_SECONDS>(**connection) >= 0 ? std::get<connectionIndexes::MIN_WAITING_TIME_SECONDS>(**connection) : parameters.getMinWaitingTimeSeconds();
 
-              if (!params.returnAllNodesResult && !reachedAtLeastOneAccessNode && nodesAccessTravelTime[nodeDepartureIndex] != -1) // check if the departure node is accessable
+              auto nodeDeparatureInNodesAccessIte = nodesAccess.find(nodeDeparture.uid);
```
tahini: typo, nodeDeparture*. Should it rather be departureNode, because it's a Node, right?

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -138,49 +139,55 @@ namespace TrRouting
...
                   }
                   if (
-                    nodeDepartureIndex == transferableNodeIndex
+                    nodeDeparture.uuid == transferableNode.node.uuid
                     && 
                     (
-                      std::get<journeyStepIndexes::TRANSFER_TRAVEL_TIME>(reverseAccessJourneysSteps[transferableNodeIndex]) == -1 
-                      || 
-                      std::get<connectionIndexes::TIME_DEP>(*reverseConnections[std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(reverseAccessJourneysSteps[transferableNodeIndex])].get()) <= connectionDepartureTime - connectionMinWaitingTimeSeconds
+                     //TODO Really not sure this is equivalent
```
tahini, line 168: It looks ok...

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -110,18 +111,18 @@ namespace TrRouting
...
         {
-          journey.push_front(std::make_tuple(-1,-1,-1,-1,nodesAccessTravelTime[resultingNodeIndex],-1,nodesAccessDistance[resultingNodeIndex]));
+          journey.push_front(std::make_tuple(-1,-1,-1,nodesAccess.at(resultingNode.uid).time,-1,nodesAccess.at(resultingNode.uid).distance));
         }
-        journey.push_back(std::make_tuple(-1,-1,-1,-1,nodesEgressTravelTime[bestEgressNodeIndex],-1,nodesEgressDistance[bestEgressNodeIndex]));
-
+        journey.push_back(std::make_tuple(-1,-1,-1,nodesEgress.at(bestEgressNode.value().get().uid).time,-1,nodesEgress.at(bestEgressNode.value().get().uid).distance));
 
 
+        //TODO This optimizeCaseStr string does nothing, should be removed
```
tahini, line 125: Question for @kaligrafy. It would be informative... if it was used

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: I was a bit inconsistant with the ordering of SomethingNode or NodeSomething :/

include/parameters.hpp
```diff
@@ -217,8 +218,7 @@ namespace TrRouting
         std::vector<std::unique_ptr<Scenario>> &scenarios,
         std::map<boost::uuids::uuid, int> &odTripIndexesByUuid,
         std::vector<std::unique_ptr<OdTrip>> &odTrips,
-        std::map<boost::uuids::uuid, int> &nodeIndexesByUuid,
-        std::vector<std::unique_ptr<Node>> &nodes,
+                             const std::map<boost::uuids::uuid,Node> &nodes,
```
tahini, line 221: align with previous line

connection_scan_algorithm/src/reverse_journey.cpp
kaligrafy: optimizeCase is used at the end to make better transfer, but it does not change the calculation.

kaligrafy: voir ici: https://github.com/chairemobilite/trRouting/tree/v2c/references

connection_scan_algorithm/src/initializations.cpp
greenscientist: The *TravelTime/*Distance where combined in NodeTimeDistance objets in the nodesAccess and nodesEgress object. 

We don't need to do the resize for the other since we not index with an id in an map instead of a position in a vector

include/path.hpp
```diff
@@ -16,24 +17,45 @@ namespace TrRouting
...
+         const std::vector<NodeTimeDistance> &anodesTimeDistance,
+         const std::vector<int> &atripsIdx):
+      uuid(auuid),
+      line(aline),
+      direction(adirection),
+      internalId(ainternalId),
+      tripsIdx(atripsIdx)
+      {
+        for (const NodeTimeDistance & ntd: anodesTimeDistance) {
+          nodesRef.push_back(ntd.node);
```
tahini, line 48: Should those 3 vectors be NodeTimeDistance, to avoid index mismatch for some unexplained reason...

tahini: If so, you don't have to do it now, just add a TODO

greenscientist: Maybe, we would need to check how they are used.

connection_scan_algorithm/src/reverse_journey.cpp
greenscientist: The optimiseCase variable yes, but the optimizeCasesStr is not used

src/od_trips_cache_fetcher.cpp
```diff
@@ -146,27 +146,23 @@ namespace TrRouting
...
-            std::vector<int> originNodesDistancesMeters(originNodesCount);
+            std::vector<NodeTimeDistance> originNodes;
             for (int i = 0; i < originNodesCount; i++)
             {
               std::string nodeUuid {capnpT.getOriginNodesUuids()[i]};
-              originNodesIdx               [i] = nodeIndexesByUuid.at(uuidGenerator(nodeUuid));
-              originNodesTravelTimesSeconds[i] = capnpT.getOriginNodesTravelTimes()[i];
-              originNodesDistancesMeters   [i] = capnpT.getOriginNodesDistances()[i];
+              originNodes.push_back(NodeTimeDistance(nodes.at(uuidGenerator(nodeUuid)),
+                                                     capnpT.getOriginNodesTravelTimes()[i],
```
tahini, line 154: alignment

src/od_trips_cache_fetcher.cpp
```diff
@@ -146,27 +146,23 @@ namespace TrRouting
...
-            std::vector<int> destinationNodesDistancesMeters(destinationNodesCount);
+            std::vector<NodeTimeDistance> destinationNodes;
             for (int i = 0; i < destinationNodesCount; i++)
             {
               std::string nodeUuid {capnpT.getDestinationNodesUuids()[i]};
-              destinationNodesIdx               [i] = nodeIndexesByUuid.at(uuidGenerator(nodeUuid));
-              destinationNodesTravelTimesSeconds[i] = capnpT.getDestinationNodesTravelTimes()[i];
-              destinationNodesDistancesMeters   [i] = capnpT.getDestinationNodesDistances()[i];
+              destinationNodes.push_back(NodeTimeDistance(nodes.at(uuidGenerator(nodeUuid)),
+                                                          capnpT.getDestinationNodesTravelTimes()[i],
```
tahini, line 164: align

src/od_trips_cache_fetcher.cpp
greenscientist: alignment is ok here, it's not in the NodeTimeDistance object, so aligning with opening (

src/od_trips_cache_fetcher.cpp
greenscientist: same

src/trips_and_connections_cache_fetcher.cpp
```diff
@@ -117,14 +113,11 @@ namespace TrRouting
               std::vector<std::unique_ptr<int>>   connectionDepartureTimes = std::vector<std::unique_ptr<int>>(nodeTimesCount);
               std::vector<std::unique_ptr<float>> connectionDemands        = std::vector<std::unique_ptr<float>>(nodeTimesCount);
 
-              for (int nodeTimeI = 0; nodeTimeI < nodeTimesCount; nodeTimeI++)
+              for (int nodeTimeI = 0; nodeTimeI < nodeTimesCount - 1; nodeTimeI++)
```
tahini, line 117: Why is this -1 necessary now and it wasn't before? Was there a bug before?

src/trips_and_connections_cache_fetcher.cpp
```diff
@@ -117,14 +113,11 @@ namespace TrRouting
...
               {
-                if (nodeTimeI < nodeTimesCount - 1)
```
tahini, line 122: oh right... this if... dismiss my previous comment

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: The previous variable where called nodeDeparture, nodeArrival, I did not want to change everything.

src/trips_and_connections_cache_fetcher.cpp
greenscientist: I did ask myself the same question when I was reviewing the code :)

tahini (approved):
It looks all good... can't say with 100% confidence, it's hard to review, but it all seems ok

greenscientist:
> It looks all good... can't say with 100% confidence, it's hard to review, but it all seems ok

Thanks, yeah, this was a bit messy, sorry

PR #185, opened 2022-10-26
Remove libtools M4 macrios from git

### LINKED ISSUE #146: libtool.m4 et ltversion.m4 are changed when compiling on macos
Should they be changed and if yes, should they be kept in the git repo or ignored?

greenscientist:


greenscientist:
Yeah, that's what I thought. Probably simply need to remove those files from the repo.

greenscientist:
This is better, removed the libtool macros, they are added by autoreconf -i anyway

tahini (approved):

PR #186, opened 2022-10-28
Migrate Scenario vector and index map to a single UUID:Scenario map

greenscientist:
Not too many changes at this point for that one

### DIFF
--- a/connection_scan_algorithm/src/route_parameters.cpp
+++ b/connection_scan_algorithm/src/route_parameters.cpp
@@ -17,7 +17,7 @@ namespace TrRouting
 
   RouteParameters::RouteParameters(std::unique_ptr<Point> orig_,
     std::unique_ptr<Point> dest_,
-    Scenario& scenario_,
+    const Scenario& scenario_,
     int _timeOfTrip,
     int minWaitingTime,
     int maxTotalTime,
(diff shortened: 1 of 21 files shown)


connection_scan_algorithm/src/route_parameters.cpp
```diff
@@ -271,7 +272,7 @@ namespace TrRouting
 
     return RouteParameters(std::make_unique<TrRouting::Point>(origin->latitude, origin->longitude),
       std::make_unique<TrRouting::Point>(destination->latitude, destination->longitude),
-      *scenario,
+                           scenario.value().get(),
```
tahini: alignment here

tests/connection_scan_algorithm/route_parameters_test.cpp
```diff
@@ -28,28 +27,21 @@ class RouteParametersFixtureTests : public BaseParametersFixtureTests
 
         // Create a valid scenario, services and other data don't have to exist, we are creating parameters, not using them
         std::unique_ptr<TrRouting::Scenario> scenario = std::make_unique<TrRouting::Scenario>();
```
tahini: Do you still need this variable here?

tahini (approved):
That's much easier to review than the node one! :D

tests/connection_scan_algorithm/route_parameters_test.cpp
greenscientist: nope, good point. (Soon, the clean up of -Wall will allow us to see this

PR #187, opened 2022-10-28
Migrate Person vector and index map to a single UUID:Person map

greenscientist:
Adjusted the logic in od_trip_routing to behave correctly when we don't have a person associated.

### DIFF
--- a/connection_scan_algorithm/src/od_trips_routing.cpp
+++ b/connection_scan_algorithm/src/od_trips_routing.cpp
@@ -236,9 +236,10 @@ namespace TrRouting
       atLeastOneCompatiblePeriod = false;
 
       // verify that od trip matches selected attributes:
-      if ( (params.odTripsAgeGroups.size()   > 0 && std::find(params.odTripsAgeGroups.begin(),   params.odTripsAgeGroups.end(),   persons[odTrip->personIdx]->ageGroup)   == params.odTripsAgeGroups.end())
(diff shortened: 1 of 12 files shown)


connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -236,9 +236,10 @@ namespace TrRouting
       atLeastOneCompatiblePeriod = false;
 
       // verify that od trip matches selected attributes:
-      if ( (params.odTripsAgeGroups.size()   > 0 && std::find(params.odTripsAgeGroups.begin(),   params.odTripsAgeGroups.end(),   persons[odTrip->personIdx]->ageGroup)   == params.odTripsAgeGroups.end())
-        || (params.odTripsGenders.size()     > 0 && std::find(params.odTripsGenders.begin(),     params.odTripsGenders.end(),     persons[odTrip->personIdx]->gender)     == params.odTripsGenders.end())
-        || (params.odTripsOccupations.size() > 0 && std::find(params.odTripsOccupations.begin(), params.odTripsOccupations.end(), persons[odTrip->personIdx]->occupation) == params.odTripsOccupations.end())
+      // TODO Find a way to simplify this if
```
tahini, line 239: omg indeed! This is one real complex if statement! Just extracting to a function would be a first step... later of course

tahini (approved):
Looks good

PR #188, opened 2022-11-02
Migrate Odtrip vector and index map to a single UUID:OdTrip map

greenscientist:
We changed the name of the global odTrip variable to odTripGlob to highlight which one was global and which was local, we had some clashes. We should aim to remove the global one in the future


tahini (approved):
Looks good I guess

PR #191, opened 2022-11-08
First pass at cleaning up the warnings

greenscientist:
This serie clean up almost all warning flagged by Wall. 

A few sign/unsigne comparison remains which require more thoughts. 

Future pass will all more warnings flags.


tahini (approved):
Looks sound, and I guess the compiler already reviewed!

PR #193, opened 2022-11-08
Upgrade to latest json.hpp

greenscientist:
Taken from https://github.com/nlohmann/json, we update from 3.7.3 to 3.11.2

This fixes several compiler warnings


tahini (approved):
Are tests sufficient to make sure the produced json is the same as before? Did you test with Transition?

greenscientist:
I was hoping the test were sufficient, but I don't really know, you are the master of the test coverage :)

greenscientist:
Tested with transition (routing and accessiblity map) and all seems good.

PR #194, opened 2022-11-09
Fix warnings: A few more misc warnings

greenscientist:
Mostly those triggered by -Wshadow -Wswitch-default, but a few others

### DIFF
--- a/connection_scan_algorithm/src/alternatives_routing.cpp
+++ b/connection_scan_algorithm/src/alternatives_routing.cpp
@@ -10,11 +10,9 @@
 #include "routing_result.hpp"
 #include "point.hpp"
 
-namespace TrRouting
-{
-
-
-  std::string LinesToString(const std::vector<std::reference_wrapper<const Line>> & linesList) {
+namespace {
+  // Placed in anynymous namespace so it's local to this file
+  std::string LinesToString(const std::vector<std::reference_wrapper<const TrRouting::Line>> & linesList) {
     std::stringstream formatter;
 
     for(auto line: linesList) {
@@ -22,7 +20,10 @@ namespace TrRouting
     }
     return formatter.str();
   }
-  
+}
+
+namespace TrRouting
+{
   /**
    * @brief Visitor class to get the line ids used by a step
    * 
@@ -283,10 +284,10 @@ namespace TrRouting
       }
 
       int i {0};
-      for (auto foundLines : alreadyFoundLines)
+      for (auto flines : alreadyFoundLines)
       {          
-        spdlog::debug("{}. {} tt: ", i, LinesToString(foundLines.first),
-                      (foundLinesTravelTimeSeconds[foundLines.first] / 60));
+        spdlog::debug("{}. {} tt: ", i, LinesToString(flines.first),
+                      (foundLinesTravelTimeSeconds[flines.first] / 60));
         i++;          
       }
       
--- a/connection_scan_algorithm/src/optimize_journey.cpp
+++ b/connection_scan_algorithm/src/optimize_journey.cpp
@@ -11,10 +11,10 @@ namespace TrRouting
   std::vector<int> Calculator::optimizeJourney(std::deque<JourneyStep> &journey)
   {
 
-    CalculationTime algorithmCalculationTime = CalculationTime();
-    algorithmCalculationTime.start();
-    long long       calculationTime;
-    calculationTime = algorithmCalculationTime.getDurationMicrosecondsNoStop();
+    CalculationTime optimizeJourneyCalculationTime = CalculationTime();
+    optimizeJourneyCalculationTime.start();
+    long long       optimizeJourneyStartCalculationTime;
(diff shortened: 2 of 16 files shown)


connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -10,19 +10,20 @@
 #include "routing_result.hpp"
 #include "point.hpp"
 
-namespace TrRouting
```
tahini, line 13: Why remove the namespace name?

tahini: oh, right, I just noticed the comment below...

connection_scan_algorithm/src/optimize_journey.cpp
```diff
@@ -11,10 +11,10 @@ namespace TrRouting
   std::vector<int> Calculator::optimizeJourney(std::deque<JourneyStep> &journey)
   {
 
-    CalculationTime algorithmCalculationTime = CalculationTime();
-    algorithmCalculationTime.start();
-    long long       calculationTime;
-    calculationTime = algorithmCalculationTime.getDurationMicrosecondsNoStop();
+    CalculationTime optimizeJourneyCalculationTime = CalculationTime();
+    optimizeJourneyCalculationTime.start();
+    long long       calculationTimeCalculationTime;
```
tahini: calculationTimeCalculationTime?? Weird naming...

connection_scan_algorithm/src/parameters.cpp
```diff
@@ -308,11 +308,12 @@ namespace TrRouting
       else if (parameterWithValueVector[0] == "od_trip_uuid")
       {
         // TODO: Use a new endpoint for od trip uuid. Now we get its od and add them to parameters
-        boost::uuids::uuid odTripUuid = uuidGenerator(parameterWithValueVector[1]);
+        // TODO maybe we should not use the global variable here
```
tahini, line 311: Indeed why use the global variable instead of just renaming the previous local one?

src/data_sources_cache_fetcher.cpp
```diff
@@ -82,6 +82,9 @@ namespace TrRouting
           case dataSource::DataSource::Type::GPS_TRACES               : t.type = "gpsTraces";               break;
           case dataSource::DataSource::Type::STREET_SEGMENT_SPEEDS    : t.type = "streetSegmentSpeeds";     break;
           case dataSource::DataSource::Type::UNKNOWN                  : t.type = "unknown";                 break;
+          default:
+            throw std::range_error("Unhandled DataSource::Type");
```
tahini: Maybe add the unhandled type in the exception?

src/od_trips_cache_fetcher.cpp
```diff
@@ -42,8 +42,10 @@ namespace TrRouting
     case odTrip::OdTrip::Mode::MULTIMODAL_OTHER : str = "multimodalOther"; break;
     case odTrip::OdTrip::Mode::OTHER            : str = "other";           break;
     case odTrip::OdTrip::Mode::UNKNOWN          : str = "unknown";         break;
+    default:
+      throw std::range_error("Unhandled DataSource::Type");
```
tahini: same and in other places too

tahini (approved):
A few comments, but looks good

src/od_trips_cache_fetcher.cpp
greenscientist: And mostly change the error message to higlight the proper type

src/data_sources_cache_fetcher.cpp
greenscientist: You mean the int value?

connection_scan_algorithm/src/parameters.cpp
greenscientist: Added comment because I don't know. The global variable is used and it's the only place that it's set. (I think it was me that changed  that in a previous clean up, but would need to verify.

connection_scan_algorithm/src/optimize_journey.cpp
greenscientist: Haha, yeah.

src/data_sources_cache_fetcher.cpp
tahini: Anything that will help the person who gets this error diagnosticate the problem without having to re-run and/or re=compile and output something... The data source ID also maybe?

tahini: I just hate generic exceptions that happen on a loop of many elements... WHY WON'T YOU TELL ME WHICH ONE CAUSED THE PROBLEM!!!  (*developer frustration*)

greenscientist: I'll put the object uuid. Ther enum type does not interact well with std::string

greenscientist:
Ok, added a few TODO to the exception message. capnp was annoying.

PR #195, opened 2022-11-09
Add -Wall and -Wextra to build flags by default

### LINKED ISSUE #190: Remove warning from builds and add -Wall by default

greenscientist:
We still have a few warnings with those, so not enable -Werror yet


tahini (approved):
Yé!

PR #196, opened 2022-11-16
Migrate Transit data storage outside of the Calculator class

greenscientist:
Create a TransitData class. This interact with a data fetcher to get and store the base transit information. All data access with this are const, so no chances to change them by the calculation.

Effectively split "constant" data and "temporary" variable. (Might have a few other variable that would be global and can be "cached"/"pre-computed")

Created a TestDataFetcher in the CSA test directory to setup all the mocked data


greenscientist:
haha, the file were masqued by the .gitignore,  so I did not see them missing in git status

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -94,92 +71,23 @@ namespace TrRouting
...
     Parameters params;
     CalculationTime algorithmCalculationTime;
-    DataFetcher &dataFetcher;
 
     // TODO Added Glob suffix to easily track which one was local and which was global
     std::optional<std::reference_wrapper<const OdTrip>> odTripGlob;
 
   private:
 
+    //TODO set it mutable so it can be changed/reset?
```
tahini, line 86: I'm starting the review here, but is it thinkable that there could be one transitData object per scenario? In which case, it could be part of the parameters instead of calculator, with one that encompass all the data. Now that it's a class, internally, transitdata can do whatever it needs.

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -94,92 +71,23 @@ namespace TrRouting
...
-    int                    updateStationsFromCache   (std::string customPath = "");
-    int                    updateAgenciesFromCache   (std::string customPath = "");
-    int                    updateServicesFromCache   (std::string customPath = "");
-    int                    updateNodesFromCache      (std::string customPath = "");
-    int                    updateLinesFromCache      (std::string customPath = "");
-    int                    updatePathsFromCache      (std::string customPath = "");
-    int                    updateScenariosFromCache  (std::string customPath = "");
-    int                    updateSchedulesFromCache  (std::string customPath = "");
-
+    //TODO migrate to TransitData
```
tahini: Why not do it now? Would that simply make this PR bigger than it should or does it have additional difficulties?

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -208,8 +116,8 @@ namespace TrRouting
 
     std::vector<NodeTimeDistance> accessFootpaths; // pair: accessNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters
     std::vector<NodeTimeDistance> egressFootpaths; // pair: egressNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters
-    std::vector<std::shared_ptr<ConnectionTuple>> forwardConnections; // Forward connections, sorted by departure time ascending
-    std::vector<std::shared_ptr<ConnectionTuple>> reverseConnections; // Reverse connections, sorted by arrival time descending
+    ////std::vector<std::shared_ptr<ConnectionTuple>> forwardConnections; // Forward connections, sorted by departure time ascending
```
tahini: remove?

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -125,10 +126,15 @@ int main(int argc, char** argv) {
     exit(-2);
   }
 
-  Calculator calculator(*fetcher);
   spdlog::info("preparing calculator...");
-  Calculator::DataStatus dataStatus = calculator.prepare();
-
+  TransitData transitData(*fetcher);
+  //TODO We wanted to handle error in the constructor, but later part of this code expect a dataStatus
```
tahini, line 131: We still want trRouting to start even if the data is invalid or has errors. The cache can be reloaded for instance (like in the simulations) so the transitData content may change. Anyway, constructors throwing errors is usually bad, but you could use a factory method to create it if you'd rather not have an "faulty" TransitData object. Anyway, current approach is good for now.

tahini (approved):
I'm glad to see this! Some questions, but this looks good otherwise. It's been tested with Transition? Any impact on performance?

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: I have to disagree here. Throwing in a constructor, if the object is invalid is usually the right thing to do. 
The debate here is what do we consider a valid state for TransitData.

connection_scan_algorithm/include/calculator.hpp
greenscientist: Currently the list of scenario is part of the TransitData object. Be seeing how it's used in the test, it might be an idea to move it out. But that would also imply changing it's location in the cache files. This will be a larger discussion.

connection_scan_algorithm/include/calculator.hpp
greenscientist: Hum I think I just need to remove it.

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: But there will be place to review the whole data flow around this. There's a few edge case, like reloading partial part of the data, that could be better handled. (And there's implication in the thread safetyness of the data. Thinking this might be a good use case for RCU, but we'll see...)

greenscientist:
The performance impact, is on par with the latest code, it does not change much.

greenscientist:
confirm working with transition

PR #202, opened 2022-11-19
Update docker build to use bullseye

greenscientist:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM debian:buster-slim 
+FROM debian:bullseye-slim
```
tahini, line 1: ha! so there was an extra white space before! :p

tahini (approved):

PR #203, opened 2022-11-19
Rework Connection management

greenscientist:
We create an actual Connection class instead of relying on a tuple. We migrate some usage to use a reference instead of a vector index. 

For the moment we still use a shared_ptr for the Connection, but we could later migrate them directly to references.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -37,7 +37,7 @@ namespace TrRouting
   class AlternativesResult;
   class TransitData;
 
-  using JourneyStep = std::tuple<int,int,std::optional<std::reference_wrapper<const Trip>>,int,short,int>; //final enter connection, final exit connection, final trip, transfer travel time, is same node transfer (first, second, third and fourth values = -1 for access and egress journeys)
+  using JourneyStep = std::tuple<std::optional<std::shared_ptr<Connection>>,std::optional<std::shared_ptr<Connection>>,std::optional<std::reference_wrapper<const Trip>>,int,short,int>; //final enter connection, final exit connection, final trip, transfer travel time, is same node transfer (first, second, third and fourth values = -1 for access and egress journeys)
 
   class Calculator {
 
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -100,8 +100,9 @@ namespace TrRouting
     long long        calculationTime;
     std::string      accessMode;
     std::string      egressMode;
-    std::vector<int> forwardConnectionsIndexPerDepartureTimeHour;
-    std::vector<int> reverseConnectionsIndexPerArrivalTimeHour;
+    //TODO
+    ////std::vector<int> forwardConnectionsIndexPerDepartureTimeHour;    
+    ////std::vector<int> reverseConnectionsIndexPerArrivalTimeHour;
 
     std::unordered_map<Node::uid_t, int> nodesTentativeTime; // arrival time at node
     std::unordered_map<Node::uid_t, int> nodesReverseTentativeTime; // departure time at node
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -100,9 +100,6 @@ namespace TrRouting
     long long        calculationTime;
     std::string      accessMode;
     std::string      egressMode;
-    //TODO
-    ////std::vector<int> forwardConnectionsIndexPerDepartureTimeHour;    
-    ////std::vector<int> reverseConnectionsIndexPerArrivalTimeHour;
 
     std::unordered_map<Node::uid_t, int> nodesTentativeTime; // arrival time at node
     std::unordered_map<Node::uid_t, int> nodesReverseTentativeTime; // departure time at node
--- a/connection_scan_algorithm/src/optimize_journey.cpp
+++ b/connection_scan_algorithm/src/optimize_journey.cpp
@@ -25,11 +25,6 @@ namespace TrRouting
     bool  startedOptimization {false};
     std::vector<std::reference_wrapper<const Node>> ignoreOptimizationNodes;
 
-    // Get a local reference to not have to call the function everywhere
-    // TODO lot of these use case could go away if we don't refer to connect by index only
-    auto & forwardConnections = transitData.getForwardConnections();
-    auto & reverseConnections = transitData.getReverseConnections();
-    
     //TODO startedOptimization can be replaced with a do...while
     while(!startedOptimization || optimizationCase >= 0)
     {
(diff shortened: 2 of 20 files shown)


greenscientist:
(Tested with Transition)

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -100,8 +100,9 @@ namespace TrRouting
     long long        calculationTime;
     std::string      accessMode;
     std::string      egressMode;
-    std::vector<int> forwardConnectionsIndexPerDepartureTimeHour;
-    std::vector<int> reverseConnectionsIndexPerArrivalTimeHour;
+    //TODO
```
tahini: what is todo?

connection_scan_algorithm/src/optimize_journey.cpp
```diff
@@ -49,31 +44,33 @@ namespace TrRouting
...
 
-          const Node & firstNodeByJourneyStep  =  std::get<connectionIndexes::NODE_DEP>(enterConnect).get();
+          //TODO Double check this, unsure what type is enterConnect
+          const Node & firstNodeByJourneyStep  =  enterConnect->getDepartureNode();
 
-          auto exitConnect = *(reverseConnections[std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION >(journeyStep)]);
+          auto exitConnect = *std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION >(journeyStep);
 
-          lastNodeByJourneyStepIdx.push_back(std::get<connectionIndexes::NODE_ARR>(exitConnect).get());
+          //TODO Double check this, unsure what type is exitConnect
```
tahini, line 63: Is this OK? Now it should always work?

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -161,7 +154,8 @@ namespace TrRouting
                   {
                     footpathDistance = nodeDeparture.reverseTransferableNodes.at(footpathIndex).distance;
                     nodesReverseTentativeTime[transferableNode.node.uid] = connectionDepartureTime - footpathTravelTime - connectionMinWaitingTimeSeconds;
-                    reverseJourneysSteps[transferableNode.node.uid] = std::make_tuple(i, currentTripQueryOverlay.exitConnection, std::cref(trip), footpathTravelTime, (nodeDeparture.uuid == transferableNode.node.uuid ? 1 : -1), footpathDistance);
+                    //TODO Do we need a make_optional<...>(connection) ??
```
tahini, line 157: I don't know... Do you need an answer from @kaligrafy?

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -350,9 +348,11 @@ namespace TrRouting
 
         if (params.returnAllNodesResult)
         {
-          if (std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(reverseAccessJourneysSteps.at(resultingNode.uid)) != -1)
+          if (std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(reverseAccessJourneysSteps.at(resultingNode.uid)).has_value())
           {
-            departureTime = std::get<connectionIndexes::TIME_DEP>(*reverseConnections[std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(reverseAccessJourneysSteps.at(resultingNode.uid))].get()) - std::get<connectionIndexes::MIN_WAITING_TIME_SECONDS>(*reverseConnections[std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(reverseAccessJourneysSteps.at(resultingNode.uid))].get());
+            //TODO Should check taht MinWaiting is not -1 or use OrDefault(0)
```
tahini, line 353: can you call the getMinWaitingTimeOrDefault as in many places above?

include/transit_data.hpp
```diff
@@ -114,8 +117,13 @@ namespace TrRouting {
...
     std::map<boost::uuids::uuid, Trip>       trips;
 
-    std::vector<std::shared_ptr<ConnectionTuple>> forwardConnections; // Forward connections, sorted by departure time ascending
-    std::vector<std::shared_ptr<ConnectionTuple>> reverseConnections; // Reverse connections, sorted by arrival time descending
+    std::vector<std::shared_ptr<Connection>> forwardConnections; // Forward connections, sorted by departure time ascending
+    std::vector<std::shared_ptr<Connection>> reverseConnections; // Reverse connections, sorted by arrival time descending
+
+    // Contains iterator matching each hour of the day from the corresponding connections container.
+    // Used to speed up iterating the connections by skipping the connections that are too early or too late
+    std::vector<std::vector<std::shared_ptr<Connection>>::const_iterator> forwardConnectionsBeginIteratorCache;
```
tahini, line 125: iiuc, those are specific to the current query, right? If so, maybe add a todo to move them to calculation specific classes

src/transit_data.cpp
```diff
@@ -135,16 +135,62 @@ namespace TrRouting {
...
+      return reverseConnections.cend();
+    }
+
+    return reverseConnectionsBeginIteratorCache[hour];
+  };
+
+  // Create a cache with a begin iterator which match the connection closest to the specified hour
+  void TransitData::generateConnectionsIteratorCache() {
+    int currentHour = 0;
+    for (std::vector<std::shared_ptr<Connection>>::const_iterator ite = forwardConnections.cbegin();
```
tahini, line 166: I'm not fully sure I understand the while inside the for, could you comment the role of each loop here?

tahini (commented):
Some comments and questions, I'll test it locally before approving

include/transit_data.hpp
greenscientist: No, this is specific to the Connection data, which is all based on the cache, the query have no impact on this. (Hence putting it in TransitData

connection_scan_algorithm/src/reverse_journey.cpp
greenscientist: Don't think so, it's not the same operation. (It's not a or, it's a substraction. 
Or did I miss something and the result would be the same?

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: actually, this TODO will get taken care in the next PR.

connection_scan_algorithm/include/calculator.hpp
greenscientist: Removing the next lines that are commented out...

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: (When creating an actual class instead of a tuple, the compiler is better at matching the types and warning accordingly

connection_scan_algorithm/src/optimize_journey.cpp
greenscientist: Yeah, it will be ok, once we move the JourneyStep type to a class (next PR)

src/transit_data.cpp
greenscientist: add comments. (Took me a while to understand it in the old code, I confess)

tahini: Thank you, with the comment, it now seems obvious, which wasn't the case before!

tahini (approved):

PR #205, opened 2022-11-21
Move JourneyStep tuple to an actual class

greenscientist:
(Only the last commit for this PR, it's based on 2 others)

Test with transition. Perf similar to previous commit.

### DIFF
--- a/connection_scan_algorithm/include/journey_step.hpp
+++ b/connection_scan_algorithm/include/journey_step.hpp
@@ -1,0 +1,57 @@
+#ifndef TR_JOURNEY_STEP
+#define TR_JOURNEY_STEP
+
+namespace TrRouting {
+
+  class JourneyStep {
+  public:
+    JourneyStep(std::optional<std::shared_ptr<Connection>> _finalEnterConnection,
+                std::optional<std::shared_ptr<Connection>> _finalExitConnection,
+                std::optional<std::reference_wrapper<const Trip>> _finalTrip,
+                int _transferTravelTime,
+                bool _sameNodeTransfer,
+                int _transferDistance) : finalEnterConnection(_finalEnterConnection),
+                                         finalExitConnection(_finalExitConnection),
+                                         finalTrip(_finalTrip),
+                                         transferTravelTime(_transferTravelTime),
+                                         sameNodeTransfer(_sameNodeTransfer),
+                                         transferDistance(_transferDistance) { }
+
+    std::optional<std::shared_ptr<Connection>> getFinalEnterConnection() const {return finalEnterConnection;}
+    std::optional<std::shared_ptr<Connection>> getFinalExitConnection() const {return finalExitConnection;}
+    std::optional<std::reference_wrapper<const Trip>> getFinalTrip() const {return finalTrip;}
+    int getTransferTravelTime() const {return transferTravelTime;}
+    bool isSameNodeTransfer() const {return sameNodeTransfer;} //TODO Is is used??
+    int getTransferDistance() const {return transferDistance;}
+
+    void setFinalEnterConnection(std::shared_ptr<Connection> _connection) {
+      finalEnterConnection = _connection;
+    }
+    void setFinalExitConnection(std::shared_ptr<Connection> _connection) {
+      finalExitConnection = _connection;
+    }
+
+    void setTransferTimeDistance(int _transferTime, int _transferDistance) {
+      transferTravelTime = _transferTime;
+      transferDistance = _transferDistance;
+    }
+    // Copy time and distance from another JourneyStep object
+    void copyTransferTimeDistance(const JourneyStep &_other) {
+      transferTravelTime = _other.transferTravelTime;
+      transferDistance = _other.transferDistance;
+    }
+
+    // Return true if both final and exit connections are valid
+    bool hasConnections() const {return finalEnterConnection.has_value() && finalExitConnection.has_value();}
+  private:
+    std::optional<std::shared_ptr<Connection>> finalEnterConnection;
+    std::optional<std::shared_ptr<Connection>> finalExitConnection;
+    std::optional<std::reference_wrapper<const Trip>> finalTrip;
+    int transferTravelTime;
+    bool sameNodeTransfer;
+    int transferDistance;
+  };
+
+}
+
+#endif
--- a/connection_scan_algorithm/include/journey_step.hpp
+++ b/connection_scan_algorithm/include/journey_step.hpp
@@ -3,6 +3,11 @@
 
 namespace TrRouting {
 
+  //a journey step can be walking from origin to first stop,
+  //riding between boarding and unboarding stops in vehicle,
+  //transfering by foot or walking from last stop to destination
+  //TODO We could change this to be a base class and have derived one for
+  //each of the step types. See issue #209
   class JourneyStep {
   public:
     JourneyStep(std::optional<std::shared_ptr<Connection>> _finalEnterConnection,
@@ -17,13 +22,22 @@ namespace TrRouting {
                                          sameNodeTransfer(_sameNodeTransfer),
                                          transferDistance(_transferDistance) { }
 
+    //TODO Why is there Final in the function name. Is that appropriate?
     std::optional<std::shared_ptr<Connection>> getFinalEnterConnection() const {return finalEnterConnection;}
     std::optional<std::shared_ptr<Connection>> getFinalExitConnection() const {return finalExitConnection;}
     std::optional<std::reference_wrapper<const Trip>> getFinalTrip() const {return finalTrip;}
+    //TODO: Comment from @tahini.
+    // So walking is tranferring in a transfer step, but walking is just walking in an access/egress step.
+    // So for the function name, getWalkingTravelTime or getWalkingTravelDistance would be more appropriate
+    //(since transfer is a concept that means something that isn't this). Or anything that means what it
+    //should and does not imply transferring from one line to the other.
+    // Again, this is something that might be more meaningful after a refactoring #209
     int getTransferTravelTime() const {return transferTravelTime;}
     bool isSameNodeTransfer() const {return sameNodeTransfer;} //TODO Is is used??
     int getTransferDistance() const {return transferDistance;}
 
+    //TODO Setting connections midway should also affected other parameters automatically
+    // Their usage in the optimise function should be reviewed (Maybe in conjunction with #209)
     void setFinalEnterConnection(std::shared_ptr<Connection> _connection) {
       finalEnterConnection = _connection;
     }
@@ -42,6 +56,9 @@ namespace TrRouting {
     }
 
     // Return true if both final and exit connections are valid
+    // TODO This could have a more meaningful name, if we find out what this represent in real life
+    // In one instance, the comment associated with calling it says "check if it is an in-vehicle journey"
+    // If we have type of each step type, maybe this would be needed. #209
     bool hasConnections() const {return finalEnterConnection.has_value() && finalExitConnection.has_value();}
   private:
     std::optional<std::shared_ptr<Connection>> finalEnterConnection;
--- a/connection_scan_algorithm/src/forward_calculation.cpp
+++ b/connection_scan_algorithm/src/forward_calculation.cpp
@@ -78,7 +78,7 @@ namespace TrRouting
           nodeWasAccessedFromOrigin  = parameters.getMaxFirstWaitingTimeSeconds() > 0 &&
             nodesAccessIte != nodesAccess.end() &&
             nodesAccessIte->second.time >= 0 &&
-            !(std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(forwardJourneysSteps.at(nodeDeparture.uid))).has_value();
+            !forwardJourneysSteps.at(nodeDeparture.uid).getFinalEnterConnection().has_value();
 
           // reachable connections only here:
           if (
@@ -117,7 +117,7 @@ namespace TrRouting
             {
               currentTripQueryOverlay.usable = true;
               currentTripQueryOverlay.enterConnection = *connection;
-              currentTripQueryOverlay.enterConnectionTransferTravelTime = std::get<journeyStepIndexes::TRANSFER_TRAVEL_TIME>(forwardJourneysSteps.at(nodeDeparture.uid));
+              currentTripQueryOverlay.enterConnectionTransferTravelTime = forwardJourneysSteps.at(nodeDeparture.uid).getTransferTravelTime();
             }
             
             if ((**connection).canUnboard() && currentTripQueryOverlay.enterConnection.has_value())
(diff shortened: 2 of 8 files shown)


greenscientist:
rebased

connection_scan_algorithm/include/journey_step.hpp
```diff
@@ -0,0 +1,57 @@
+#ifndef TR_JOURNEY_STEP
+#define TR_JOURNEY_STEP
+
+namespace TrRouting {
+
+  class JourneyStep {
```
tahini, line 11: What does this class represent? Since enter and exit connections are optional, it can be any journey step, like access, egress, transfer, in vehicle, etc? Or does it imply a tranfers? Since there are transfer travel time and distances fields...

connection_scan_algorithm/include/journey_step.hpp
```diff
@@ -0,0 +1,57 @@
...
+                                         transferDistance(_transferDistance) { }
+
+    std::optional<std::shared_ptr<Connection>> getFinalEnterConnection() const {return finalEnterConnection;}
+    std::optional<std::shared_ptr<Connection>> getFinalExitConnection() const {return finalExitConnection;}
+    std::optional<std::reference_wrapper<const Trip>> getFinalTrip() const {return finalTrip;}
+    int getTransferTravelTime() const {return transferTravelTime;}
+    bool isSameNodeTransfer() const {return sameNodeTransfer;} //TODO Is is used??
+    int getTransferDistance() const {return transferDistance;}
+
+    void setFinalEnterConnection(std::shared_ptr<Connection> _connection) {
```
tahini, line 41: I know it's a tuple turned into a class, but would it be desirable eventually to make these objects immutable? Or are setters necessary? (I'm thinking that setting a connection will potentially impact the transfer time and distance, so they need to be set together, no?)

If I'm totally mistaken, please reply to this comment, otherwise, you can add proper todos in the code or issues.

connection_scan_algorithm/include/journey_step.hpp
```diff
@@ -0,0 +1,57 @@
...
+      transferTravelTime = _transferTime;
+      transferDistance = _transferDistance;
+    }
+    // Copy time and distance from another JourneyStep object
+    void copyTransferTimeDistance(const JourneyStep &_other) {
+      transferTravelTime = _other.transferTravelTime;
+      transferDistance = _other.transferDistance;
+    }
+
+    // Return true if both final and exit connections are valid
```
tahini, line 58: "enter* and exit". Also, by connection, you mean transfer? Doesn't an access/egress step have a connection, since at least enter or exit _connections_ are set? What does it mean to have both set vs only one?

connection_scan_algorithm/src/forward_calculation.cpp
```diff
@@ -172,12 +172,12 @@ namespace TrRouting
                      //TODO Not fully sure this is equivalent to the ancient code
                      forwardEgressJourneysSteps.count(transferableNode.node.uid) == 0
                       ||
-                     std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION>(forwardEgressJourneysSteps.at(transferableNode.node.uid)).value()->getArrivalTime() > connectionArrivalTime
+                     forwardEgressJourneysSteps.at(transferableNode.node.uid).getFinalExitConnection().value()->getArrivalTime() > connectionArrivalTime
```
tahini, line 175: Are you sure you have a value here?

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -99,18 +99,18 @@ namespace TrRouting
...
           journey.push_front(resultingNodeJourneyStep);
-          bestAccessNode = std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(resultingNodeJourneyStep).value()->getDepartureNode();
+          bestAccessNode = resultingNodeJourneyStep.getFinalEnterConnection().value()->getDepartureNode();
           resultingNodeJourneyStep = forwardJourneysSteps.at(bestAccessNode.value().get().uid);
         }
 
         if (!params.returnAllNodesResult)
         {
-          journey.push_back(std::make_tuple(std::nullopt,std::nullopt,std::nullopt,nodesEgress.at(resultingNode.uid).distance,-1,nodesEgress.at(resultingNode.uid).distance));
+          journey.push_back(JourneyStep(std::nullopt,std::nullopt,std::nullopt,nodesEgress.at(resultingNode.uid).distance,false,nodesEgress.at(resultingNode.uid).distance));
```
tahini: You could add some breathing space to those lines and add a whitespace after the commas. Same below.

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -119,17 +119,17 @@ namespace TrRouting
         for (auto & journeyStep : journey)
         {
 
-          if (std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(journeyStep).has_value() && std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION>(journeyStep).has_value())
+          if (journeyStep.hasConnections())
```
tahini, line 132: If I look at the else, `hasConnections` is more `isTransitStep` or something like that, something not access/egress.

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -140,9 +140,9 @@ namespace TrRouting
             transferArrivalTime        = arrivalTime   + transferTime;
             transferReadyTime          = transferArrivalTime;
 
-            if (journey.size() > i + 1 && std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(journey[i+1]).has_value())
+            if (journey.size() > i + 1 && journey[i+1].getFinalEnterConnection().has_value())
```
tahini, line 153: Also, why final? final in what sense? Why not just getEnterConnection?

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -265,8 +265,8 @@ namespace TrRouting
           else // access or egress journey step
           {
 
-            transferTime          = std::get<journeyStepIndexes::TRANSFER_TRAVEL_TIME>(journeyStep);
-            distance              = std::get<journeyStepIndexes::TRANSFER_DISTANCE>(journeyStep);
+            transferTime          = journeyStep.getTransferTravelTime();
```
tahini, line 278: The else says "access or egress journey step", but you get "transferTime". Would it be more "walkingTime" if you want to use it both for transfers and access/egress?

connection_scan_algorithm/src/optimize_journey.cpp
```diff
@@ -297,8 +295,8 @@ namespace TrRouting
               if (exitConnection.has_value() && connection->canBoard())
               {
                 usedOptimizationCases.push_back(4);
-                std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION>(journey[fromJourneyStepIdx]) = exitConnection;
-                std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(journey[toJourneyStepIdx])  = connection;
+                journey[fromJourneyStepIdx].setFinalExitConnection(exitConnection.value());
```
tahini, line 298: Previously, you were setting `exitConnection` and `connection`, but now, you set one with `exitConnection.value()` and the other with simply `connection`. Did the compiler miss something before or am I missing something here?

connection_scan_algorithm/src/reverse_journey.cpp
```diff
@@ -92,24 +92,23 @@ namespace TrRouting
...
           journey.push_back(resultingNodeJourneyStep);
-          bestEgressNode      = std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION>(resultingNodeJourneyStep).value()->getArrivalNode();
+          bestEgressNode = resultingNodeJourneyStep.getFinalExitConnection().value()->getArrivalNode();
           resultingNodeJourneyStep = reverseJourneysSteps.at(bestEgressNode.value().get().uid);
         }
 
         if (!params.returnAllNodesResult)
         {
-          journey.push_front(std::make_tuple(std::nullopt,std::nullopt,std::nullopt,nodesAccess.at(resultingNode.uid).time,-1,nodesAccess.at(resultingNode.uid).distance));
+          journey.push_front(JourneyStep(std::nullopt,std::nullopt,std::nullopt,nodesAccess.at(resultingNode.uid).time,false,nodesAccess.at(resultingNode.uid).distance));
```
tahini: white spaces here

tahini (changes requested):
Refactoring code focuses all the attention to that code... and its un-clarities... sorry about that...

connection_scan_algorithm/include/journey_step.hpp
greenscientist: You'll have to ask @kaligrafy for an answer like that.

connection_scan_algorithm/include/journey_step.hpp
greenscientist: No real idea. I just formalized that actions that were done in the code already.

connection_scan_algorithm/include/journey_step.hpp
greenscientist: another questin for @kaligrafy :)

connection_scan_algorithm/src/forward_calculation.cpp
greenscientist: I haven't checked the code yet, but I think it was one of the case where I was not 100% sure. At least value() throws an exception if there's no value, so we will catch it.

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Yeah, I just did not touch the previous code

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Maybe :)

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: @kaligrafy ... :)

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: @kaligrafy

connection_scan_algorithm/src/optimize_journey.cpp
greenscientist: I think before was wrong somehow

greenscientist: (The tuple don't always type match correctly...)

connection_scan_algorithm/include/journey_step.hpp
greenscientist: The setter are used in the optimise_journey code path. I don't have an high trust in that part of the code. 
(I don't think it's well covered in the current unit tests)

connection_scan_algorithm/include/journey_step.hpp
kaligrafy: a journey step can be walking from origin to first stop, riding between boarding and unboarding stops in vehicle, transfering by foot or walking from last stop to destination

connection_scan_algorithm/include/journey_step.hpp
kaligrafy: I really don't remember for setFinalEnterConnection

connection_scan_algorithm/include/journey_step.hpp
kaligrafy: no, enter connection is the connection use just afeter boarding. Exit connection is the last connection before unboarding

connection_scan_algorithm/src/forward_journey.cpp
kaligrafy: the last exit connection will be the last used connection before egress

connection_scan_algorithm/src/forward_journey.cpp
kaligrafy: yes this is walking time, but it also called transfer time

kaligrafy: Because you are walking to reach transferring stop. It should be 0 if transferring at same stop

connection_scan_algorithm/include/journey_step.hpp
tahini: @greenscientist can you add this as comment above? With potentially a TODO if you think something should be done about the journey step type heterogeneity.

connection_scan_algorithm/include/journey_step.hpp
tahini: I'm still confused here. If you think of any todos 2 to add here, please do...

connection_scan_algorithm/include/journey_step.hpp
tahini: @kaligrafy so which kind of step has connections? Only in-vehicle? Or transfers, access, egress too?

connection_scan_algorithm/src/forward_calculation.cpp
tahini: ok... anyway it's the same as before I guess

connection_scan_algorithm/src/forward_journey.cpp
tahini: @kaligrafy Could you confirm what it means to have BOTH enter and exit connection? So that `hasConnections` can have a more meaningful name

connection_scan_algorithm/src/forward_journey.cpp
tahini: But given the function name, ALL connections are final in this class, not just the last one. One of them happens to be the last one, but the step doesn't know that, it's the trip flow that will know which one. So having "final" in the function name may not be necessary? @kaligrafy?

connection_scan_algorithm/src/forward_journey.cpp
tahini: So walking is tranferring in a transfer step, but walking is just walking in an access/egress step. So for the function name, `getWalkingTravelTime` or `getWalkingTravelDistance` would be more appropriate (since transfer is a concept that means something that isn't this). Or anything that means what it should and does not imply transferring from one line to the other.

connection_scan_algorithm/include/journey_step.hpp
greenscientist: Adding comment and created #209

connection_scan_algorithm/include/journey_step.hpp
greenscientist: Agreed, Added a TODO.

connection_scan_algorithm/src/forward_calculation.cpp
greenscientist: yeah

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Splited into separate lines

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Added TODO

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Ok, that's a strong question. I just copied/pasted in as a TODO.

tahini (approved):
Ok, sounds like lot of refactoring fun in perspective!

greenscientist:
haha, yes. But at least, I think we will be getting to some easy to understand code after that!

PR #208, opened 2022-11-29
Fix assert in optimise_journey

### LINKED ISSUE #207: Assertion failed for some queries
error: Assertion failed: (trip.reverseConnectionsIdx.size() > 1 + sequenceEndIdx), function optimizeJourney, file optimize_journey.cpp, line 177.

greenscientist:
Comparison should be >= instead of just >

Closes #207

### DIFF
--- a/connection_scan_algorithm/src/optimize_journey.cpp
+++ b/connection_scan_algorithm/src/optimize_journey.cpp
@@ -171,7 +171,8 @@ namespace TrRouting
         int sequenceStartIdx = std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(journey[fromJourneyStepIdx]).value()->getSequenceInTrip() - 1;
         int sequenceEndIdx   = std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION >(journey[fromJourneyStepIdx]).value()->getSequenceInTrip() - 1;
 
-        assert(trip.reverseConnections.size() > 1 + sequenceEndIdx); // make sure sequenceIdx will be valid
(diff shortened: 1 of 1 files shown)


connection_scan_algorithm/src/optimize_journey.cpp
```diff
@@ -171,7 +171,7 @@ namespace TrRouting
         int sequenceStartIdx = std::get<journeyStepIndexes::FINAL_ENTER_CONNECTION>(journey[fromJourneyStepIdx]).value()->getSequenceInTrip() - 1;
         int sequenceEndIdx   = std::get<journeyStepIndexes::FINAL_EXIT_CONNECTION >(journey[fromJourneyStepIdx]).value()->getSequenceInTrip() - 1;
 
-        assert(trip.reverseConnections.size() > 1 + sequenceEndIdx); // make sure sequenceIdx will be valid
+        assert(trip.reverseConnections.size() >= 1 + sequenceEndIdx); // make sure sequenceIdx will be valid
```
tahini, line 175: Is that the index starting with 1? Worth commenting as it would be unintuitive to anyone...

tahini (changes requested):

connection_scan_algorithm/src/optimize_journey.cpp
greenscientist: Frankly, I'm not 100% sure, but I think I've seen this while looking at other part of the code.

greenscientist:
@tahini Let me know if you prefer with this comment.

tahini (approved):

PR #211, opened 2022-12-01
Split AllNodesResult codepath for forwardJourneyStep

greenscientist:
This create a second function forwardJourneyStepAllNodes for the params.returnAllNodesResult == true case.

After multiple pass for unused variable cleanup in the AllNodes case, we ended up with a really simple function.

We don't have to create the resultingNodes vector. In the allNodes case, we iterate over all nodes and in the single result one, we just check the bestEgressNode.

Might be possible to optimize further, so far we have only done the most evident ones

### DIFF
--- a/connection_scan_algorithm/src/forward_journey.cpp
+++ b/connection_scan_algorithm/src/forward_journey.cpp
@@ -16,46 +16,24 @@
 namespace TrRouting
 {
 
-  std::unique_ptr<RoutingResult> Calculator::forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, std::optional<std::reference_wrapper<const Node>> bestEgressNode, const std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps)
+  std::unique_ptr<SingleCalculationResult> Calculator::forwardJourneyStep(RouteParameters &parameters, std::optional<std::reference_wrapper<const Node>> bestEgressNode, const std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps)
   {
+    assert(!params.returnAllNodesResult); // Just make sure we are in the right code path
 
-    // FIXME: One code path supports both Single calculation and all nodes, we shouldn't have to deal with both of these types here
     std::unique_ptr<SingleCalculationResult> singleResult = std::make_unique<SingleCalculationResult>();
-    std::unique_ptr<AllNodesResult> allNodesResult = std::make_unique<AllNodesResult>();
 
-    int              nodesCount           {1};
-    int              reachableNodesCount  {0};
-    bool             foundLine            {false};
-
-    std::vector<std::reference_wrapper<const Node>> resultingNodes;
-    if (params.returnAllNodesResult)
+    if (!bestEgressNode.has_value())
     {
-      nodesCount     = transitData.getNodes().size();
-      // Add references to all nodes in the resultingNodes vector
-      // TODO this seems heavy. To reconsidered when we split function and consider the
-      // allNodes case separately 
-      for(auto nodeIte = transitData.getNodes().begin(); nodeIte != transitData.getNodes().end(); nodeIte++) {
-        resultingNodes.push_back(std::cref(nodeIte->second));
-      }
-    }
-    else if (bestArrivalTime < MAX_INT) // line found
-    {
-      foundLine = true;
-      resultingNodes.push_back(bestEgressNode.value());
-    } else {
       throw NoRoutingFoundException(NoRoutingReason::NO_ROUTING_FOUND);
     }
 
-    //TODO This if seems unnecessary, we throw before if this condition would be false
-    if (foundLine || params.returnAllNodesResult)
-    {
+    const Node & resultingNode = bestEgressNode.value().get();
+    if (forwardEgressJourneysSteps.count(resultingNode.uid) > 0) {
+
       std::deque<JourneyStep> journey;
 
       std::shared_ptr<Connection> journeyStepEnterConnection;
       std::shared_ptr<Connection> journeyStepExitConnection;
-      std::vector<boost::uuids::uuid>                   unboardingNodeUuids;
-      std::vector<boost::uuids::uuid>                   boardingNodeUuids;
-      std::vector<int>                                  inVehicleTravelTimesSeconds; // the in vehicle travel time for each segment
       std::vector<Leg>  legs;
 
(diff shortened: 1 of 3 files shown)


connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -16,86 +16,43 @@
...
-      int totalDistance            { 0}; int distance               {-1};
-      int inVehicleDistance        {-1}; int totalInVehicleDistance { 0}; int totalWalkingDistance   { 0};
-      int totalTransferDistance    {-1}; int accessDistance         { 0}; int egressDistance         { 0};
-      int accessWalkingTime      {-1};
-      int transferTime             {-1}; int egressWalkingTime      {-1};
-      int waitingTime              {-1}; int accessWaitingTime      {-1};
-
-      for (const Node & resultingNode : resultingNodes)
-      {
+    { ////TODO Kept the { to not have to realign everything yet
```
tahini: ;-)

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -16,86 +16,43 @@
 namespace TrRouting
 {
 
-  std::unique_ptr<RoutingResult> Calculator::forwardJourneyStep(RouteParameters &parameters, int bestArrivalTime, std::optional<std::reference_wrapper<const Node>> bestEgressNode, const std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps)
+  std::unique_ptr<SingleCalculationResult> Calculator::forwardJourneyStep(RouteParameters &parameters, std::optional<std::reference_wrapper<const Node>> bestEgressNode, const std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps)
```
tahini, line 19: Tell me if I'm wrong, but the unique_ptr return value here was because of the various possible return types. It was causing problems, later. Maybe you can now return the object instead of pointer, if that's desirable of course.

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -271,97 +216,122 @@ namespace TrRouting
...
+    spdlog::debug("-- forward result: single calculation");
 
     return singleResult;
 
   }
 
+  std::unique_ptr<AllNodesResult> Calculator::forwardJourneyStepAllNodes(RouteParameters &parameters, const std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps)
+  {
+    assert(params.returnAllNodesResult); // Just make sure we are in the right code path
+    // FIXME: One code path supports both Single calculation and all nodes, we shouldn't have to deal with both of these types here
```
tahini: You just fixed it!

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -271,97 +216,122 @@ namespace TrRouting
...
+    ////std::unique_ptr<SingleCalculationResult> singleResult = std::make_unique<SingleCalculationResult>();
```
tahini: remove

tahini (commented):
You know I like everything that reduces code complexity. I just parsed through and see nothing objectionable. When you're done with the tests, I'll test it too.

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Yeah, probably. I was unsure about the implication of removing the unique_ptr.

greenscientist:
Run well locally with transition

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: Added a TODO, removing it will only be possible when all the function are splitted and converted.

tahini: Of course, I was just mentioning because iirc, the allNodes/single calculation was the main (and only) reason to use unique_ptr, but my memory may be wrong.

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -70,298 +48,285 @@ namespace TrRouting
...
+              totalInVehicleTime         += inVehicleTime;
+              totalWaitingTime           += waitingTime;
+              if (Mode::TRANSFERABLE != journeyStepTrip.line.mode.shortname)
+                {
+                  numberOfTransfers += 1;
+                }
+              legs.push_back(std::make_tuple(std::ref(journeyStepTrip), boardingSequence - 1, unboardingSequence - 1));
+
+              if (unboardingSequence - 1 < journeyStepTrip.path.segmentsDistanceMeters.size()) // check if distances are available for this path
+                {
```
tahini: Is it code style to offset the opening bracket on the next line? The code typically has the bracket aligned with the keyword above. Maybe we should document and fix the code style once and for all for trRouting.

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -70,298 +48,285 @@ namespace TrRouting
...
-              }
+                                                                                   journeyStepTrip,
+                                                                                   unboardingSequence,
+                                                                                   unboardingSequence + 1,
+                                                                                   journeyStepNodeArrival,
+                                                                                   arrivalTime,
+                                                                                   inVehicleTime,
+                                                                                   inVehicleDistance
+                                                                                   ));
+              if (i < journeyStepsCount - 2) // if not the last transit leg
```
tahini: I was about to say that this doesn't compile in my head, but then I looked at the code below and it ended up compiling :D ... This block is worth a comment above the if, eg `// Add a transfer step if this is not the last transit leg`

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -70,298 +48,285 @@ namespace TrRouting
...
-            {
-              totalTransferWaitingTime += waitingTime;
-            }
+                {
+                  inVehicleDistance      = -1;
+                  totalDistance          = -1;
+                  totalInVehicleDistance = -1;
+                }
+
+              // TODO confirm that i == 1 is the right index to compare to
```
tahini: I think the i== 0 is the access step, this would be the first transit (leg) step

tahini (approved):
Sounds good, it seems to work. As always, a request for additional comments that you may or may not address...

connection_scan_algorithm/src/forward_journey.cpp
greenscientist: No, somehow, my emacs does that sometimes and I forgot to realign

PR #214, opened 2022-12-02
v1: Increment the alternativeSequence field in result

tahini:
This field is part of the API and is expected to be incremented for each alternative. Start at 1 because it's human readable.

### DIFF
--- a/connection_scan_algorithm/src/result_to_v1.cpp
+++ b/connection_scan_algorithm/src/result_to_v1.cpp
@@ -182,11 +182,13 @@ namespace TrRouting
     nlohmann::json json;
     json["status"] = STATUS_SUCCESS;
     json["alternatives"] = nlohmann::json::array();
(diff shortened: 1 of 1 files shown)


connection_scan_algorithm/src/result_to_v1.cpp
```diff
@@ -182,11 +182,12 @@ namespace TrRouting
     nlohmann::json json;
     json["status"] = STATUS_SUCCESS;
     json["alternatives"] = nlohmann::json::array();
-    int alternativeSequence = 0;
+    int alternativeSequence = 1;
```
greenscientist, line 186: Please add a comment stating why we start at 1 :)

greenscientist:
One could argue that json is not really intended to be human readable.... :)

tahini:
alternativeSequence is not part of v2

greenscientist (approved):

PR #215, opened 2022-12-02
v2: Add the `v2/accessibility` API route

tahini:
This route is for the all_nodes of v1, with only the required parameters, a single place and the time query parameters are the same as the other calculation routes.

The response type is defined and the node's code and name are returned additionally, like the other API responses. Consumers don't necessarily have access to the data and the ID may not be meaningful.

### DIFF
--- a/docs/APIv2/accessibilityResponse.yml
+++ b/docs/APIv2/accessibilityResponse.yml
@@ -1,0 +1,105 @@
+access_query_error: # 'query_error' is a value for the status (discriminator)
+  required:
+    - status
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [query_error]
+    errorCode:
+      type: string
+      enum:
+        - 'EMPTY_SCENARIO'
+        - 'MISSING_PARAM_SCENARIO'
+        - 'MISSING_PARAM_PLACE'
+        - 'MISSING_PARAM_TIME_OF_TRIP'
+        - 'INVALID_PLACE'
+        - 'INVALID_NUMERICAL_DATA'
+        - 'PARAM_ERROR_UNKNOWN'
+
+baseRouteResponse:
+  type: object
+  properties:
+    place:
+      type: array
+      items:
+        type: number
+      minItems: 2
+      maxItems: 2
+      description: Longitude and latitude of the requested place, in the WSG84 coordiantes system
+    timeOfTrip:
+      type: integer
+      description: |
+        The requested time of the trip, in seconds since midnight.
+        The time_type field will determine if it represents a departure or arrival time.
+    timeType:
+      type: integer
+      enum:
+        - 0
+        - 1
+      description: The type of the requestTime. 0 means it is the departure time; 1 means arrival time
+
+NoRoutingFound:
+  required:
+    - status
+  allOf:
+  - $ref: '#/baseRouteResponse'
+  - type: object
+    properties:
+      status:
+        type: string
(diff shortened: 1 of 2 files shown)


docs/APIv2/accessibilityResponse.yml
```diff
@@ -0,0 +1,102 @@
...
+  - type: object
+    properties:
+      status:
+        type: string
+        enum: [no_routing_found]
+      reason:
+        type: string
+        enum:
+          # Generic reason, when not possible to specify more
+          - NO_ROUTING_FOUND,
```
tahini, line 56: Wondering if the next 2 are enough or if there's really need for this third reason... The code will tell... But yeah, maybe if we can reach a stop, wait for the bus, enter the bus, but we don't even reach the next stop within the requested time... I guess we can leave it

docs/APIv2/accessibilityResponse.yml
```diff
@@ -0,0 +1,102 @@
...
+        description: Name of the current node
+      nodeCode:
+        type: string
+        description: Code of the current node
+      nodeUuid:
+        type: string
+        description: UUID of the current node
+      nodeTime:
+        type: number
+        description: Time at which this node was reached, in seconds since midnight. This represents the departure time from node if time is the arrival, or arrival time if time is the departure
```
greenscientist: unsure if I understand well. 

Do you mean, in the second part, if the time *type" is arrival or arrival time if time *type* is departure ?

tahini: "if the requested time" would be more exact (time type is 0 or 1 to mean departure or arrival)

greenscientist:
Seem sane to me. Will leave around for the weekend in case @kaligrafy want to have a look.

greenscientist (approved):

docs/APIv2/accessibilityResponse.yml
```diff
@@ -0,0 +1,102 @@
...
+        type: string
+        enum: [success]
+
+accessibilityResponse:
+  type: object
+  properties:
+    nodes:
+      type: array
+      items:
+        $ref: '#/nodeAccessibility'
```
tahini, line 80: Il manque le pourcentage de noeuds accessibles. On a le nombre, mais le pourcentage permet de savoir combien de nombre au total ont été testés

greenscientist: C'est une partie que je trouvais un peu abérante dans le code. Pourrait-on mettre le nombre total de noeud à la place? Le pourcentage donne un nombre un peu imprécis, on peut laisser le client calculer le pourcentage...

Sinon, il faudrait préciser la précision du nombre qu'on met ici. C'est un float, donc par définition non exact.

greenscientist (changes requested):

docs/APIv2/accessibilityResponse.yml
tahini: done

greenscientist: Do we want to add a comment saying something along "totalNode is useful to calculate the percentage of node accessible" or that's not relevent to the API doc ?

greenscientist (approved):

docs/APIv2/accessibilityResponse.yml
tahini: I don't think it's relevant for API doc, one can do whatever they want with this information!

PR #216, opened 2022-12-02
Split all nodes calculate

greenscientist:
split the calculate function

Also do a bunch of related small clean ups

(Based on previous split PR)

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -62,6 +62,9 @@ namespace TrRouting
     // use common calculation functions
     // TODO Once the split is done, we can get rid of the unique_ptr return and have the right concret type returned directly
     std::unique_ptr<RoutingResult> calculate(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
+    std::unique_ptr<SingleCalculationResult> calculateSingle(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
+    std::unique_ptr<AllNodesResult> calculateAllNodes(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
+
     // Forward and and reverse calculation, in addition to their return values will fill up their JourneysSteps map
     std::optional<std::tuple<int, std::reference_wrapper<const Node>>> forwardCalculation(RouteParameters &parameters, std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps); // best arrival time,   best egress node
     void forwardCalculationAllNodes(RouteParameters &parameters, std::unordered_map<Node::uid_t, JourneyStep> & forwardEgressJourneysSteps);
@@ -98,6 +101,7 @@ namespace TrRouting
     void resetFilters(const RouteParameters &parameters);
     // Convert the optimization case ID returned by optimizeJourney to a string
     std::string optimizeCasesToString(const std::vector<int> optimizeCases);
+    std::unique_ptr<SingleCalculationResult> calculateSingleReverse(RouteParameters &parameters);
 
     //TODO set it mutable so it can be changed/reset?
     const TransitData &transitData;
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -107,7 +107,6 @@ namespace TrRouting
     const TransitData &transitData;
 
     int              departureTimeSeconds;
-    int              initialDepartureTimeSeconds;
     int              arrivalTimeSeconds;
     int              minAccessTravelTime;
     int              maxEgressTravelTime;
--- a/connection_scan_algorithm/src/calculator.cpp
+++ b/connection_scan_algorithm/src/calculator.cpp
@@ -28,7 +28,6 @@ namespace TrRouting
     if (departureTimeSeconds > -1 && parameters.isForwardCalculation())
     {
       
-      initialDepartureTimeSeconds = departureTimeSeconds; // set initial departure time so we can find the latest possible departure time with reverse calculation later and still know the initial waiting time
       int bestArrivalTime {MAX_INT};
       std::optional<std::reference_wrapper<const Node>> bestEgressNode;
       //TODO With the TODO later that forwardJourneyStep is not necessary, we can also drop this variable
@@ -72,7 +71,6 @@ namespace TrRouting
     else if (arrivalTimeSeconds > -1)
     {
       departureTimeSeconds = -1;
-      initialDepartureTimeSeconds = -1;
       //TODO maybe we can do something different in that case, like a query flag
       // we need to make all trips usable when not coming from forward result because reverse calculation, by default, checks for usableTrips
       for (auto && tripIte : transitData.getTrips()) {
@@ -120,8 +118,6 @@ namespace TrRouting
     {
       std::unordered_map<Node::uid_t, JourneyStep> forwardEgressJourneysSteps;
 
-      initialDepartureTimeSeconds = departureTimeSeconds; // set initial departure time so we can find the latest possible departure time with reverse calculation later and still know the initial waiting time
-
       forwardCalculationAllNodes(parameters, forwardEgressJourneysSteps);
 
       spdlog::debug("-- forward calculation all nodes -- {} microseconds", algorithmCalculationTime.getDurationMicrosecondsNoStop() - calculationTime);
@@ -138,7 +134,6 @@ namespace TrRouting
       std::unordered_map<Node::uid_t, JourneyStep> reverseAccessJourneysSteps;
 
       departureTimeSeconds = -1;
-      initialDepartureTimeSeconds = -1;
       //TODO maybe we can do something different in that case, like a query flag
       // we need to make all trips usable when not coming from forward result because reverse calculation, by default, checks for usableTrips
       for (auto && tripIte : transitData.getTrips()) {
--- a/connection_scan_algorithm/src/calculator.cpp
+++ b/connection_scan_algorithm/src/calculator.cpp
@@ -13,85 +13,60 @@ namespace TrRouting
   
   std::unique_ptr<RoutingResult> Calculator::calculate(RouteParameters &parameters, bool resetAccessPaths, bool resetFilters) {
 
+    if (params.returnAllNodesResult) {
+      return calculateAllNodes(parameters, resetAccessPaths, resetFilters);
+    } else {
+      return calculateSingle(parameters, resetAccessPaths, resetFilters);
+    }
+  }
+
+  std::unique_ptr<SingleCalculationResult> Calculator::calculateSingle(RouteParameters &parameters, bool resetAccessPaths, bool resetFilters) {
     reset(parameters, resetAccessPaths, resetFilters);
 
-    std::unique_ptr<RoutingResult> result;
-    
-    int bestArrivalTime {MAX_INT};
-    int bestDepartureTime {-1};
-    std::optional<std::reference_wrapper<const Node>> bestEgressNode;
-    std::optional<std::reference_wrapper<const Node>> bestAccessNode;
-
-    // TODO Could they be local to each section?
-    std::unordered_map<Node::uid_t, JourneyStep> forwardEgressJourneysSteps;
-    std::unordered_map<Node::uid_t, JourneyStep> reverseAccessJourneysSteps;
-
+    std::unique_ptr<SingleCalculationResult> result;
 
     if (departureTimeSeconds > -1 && parameters.isForwardCalculation())
     {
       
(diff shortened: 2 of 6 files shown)


connection_scan_algorithm/src/calculator.cpp
```diff
@@ -13,24 +13,25 @@ namespace TrRouting
...
 
-    reset(parameters, resetAccessPaths, resetFilters);
-
-    std::unique_ptr<RoutingResult> result;
-    
-    int bestArrivalTime {MAX_INT};
-    int bestDepartureTime {-1};
-    std::optional<std::reference_wrapper<const Node>> bestEgressNode;
-    std::optional<std::reference_wrapper<const Node>> bestAccessNode;
+    if (params.returnAllNodesResult) {
```
tahini, line 16: Can you move this `if` to the caller of calculate, so the calculate method can be removed altogether? (And single and allNodes methods can have a different class for parameters, I'm thinking of the v2 allnodes route implementation here...)

greenscientist: Yes, but that's for future PR. This one is already big enough. 

We will probably need to fully convert everything to API v2 before we can remove the call to calculate. (There's some oddities in the OdTrip calculation that needs clean up before we can do full conversion)

tahini: do odTrip and allNodes have anything in common? The allNodes query does not have an origin and a destination, so using the RouteParameters for it is un-necessary and we could have a AccessibilityParameters class that extend a common class for common parameters (access/egress times, etc), but with only a single place Point and single time, either forward or not. The allNodes query could be completely independent from the rest and the class parameters.

Future PR, yes, but very near future. Like I can do it while implementing the v2 accessibility call if you wish and don't see anything major preventing it. Anyhow, I'll wait for this PR to do the v2 accessibility.

greenscientist: I did not want to just move the if outside of the calculate call. I prefer to just call the appropriate calculate when it's the right one. Yes, the if will be require in some part (for V1 only I think, for V2, it would be necessary at all), but for example, in alternativeRouting, We can just call calculateSingle there. 
Once that is done, we can cleanup the result processing also.

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -98,12 +101,12 @@ namespace TrRouting
     void resetFilters(const RouteParameters &parameters);
     // Convert the optimization case ID returned by optimizeJourney to a string
     std::string optimizeCasesToString(const std::vector<int> optimizeCases);
+    std::unique_ptr<SingleCalculationResult> calculateSingleReverse(RouteParameters &parameters);
```
tahini, line 104: Does this have to be in the class? Or can it just be defined locally in the calculator.cpp file? I guess it uses the class's internal fields for now...

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -13,125 +13,146 @@ namespace TrRouting
...
 
-    // TODO Could they be local to each section?
-    std::unordered_map<Node::uid_t, JourneyStep> forwardEgressJourneysSteps;
-    std::unordered_map<Node::uid_t, JourneyStep> reverseAccessJourneysSteps;
+  std::unique_ptr<SingleCalculationResult> Calculator::calculateSingle(RouteParameters &parameters, bool resetAccessPaths, bool resetFilters) {
+    reset(parameters, resetAccessPaths, resetFilters);
 
+    std::unique_ptr<SingleCalculationResult> result;
 
     if (departureTimeSeconds > -1 && parameters.isForwardCalculation())
```
tahini, line 28: I recall from a previous comment somewhere you mentioning that we could just use the `isForwardCalculation`, right?

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -13,125 +13,146 @@ namespace TrRouting
...
-          calculationTime = algorithmCalculationTime.getDurationMicrosecondsNoStop();
+        assert(false); // See TODO
+        spdlog::debug("-- forward journey -- {} microseconds", algorithmCalculationTime.getDurationMicrosecondsNoStop() - calculationTime);
+        calculationTime = algorithmCalculationTime.getDurationMicrosecondsNoStop();
           
-        }
-        
       }
     }
     else if (arrivalTimeSeconds > -1)
```
tahini, line 71: Use the `isForwardCalculation` only above makes this a simple else

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -13,125 +13,146 @@ namespace TrRouting
...
-      if (params.returnAllNodesResult) { //TODO Temporary until we split calculate()
+    }
 
-        reverseCalculationAllNodes(parameters, reverseAccessJourneysSteps);
+    return result;
+  }
 
-        spdlog::debug("-- reverse calculation --  {} microseconds", algorithmCalculationTime.getDurationMicrosecondsNoStop() - calculationTime);
-        calculationTime = algorithmCalculationTime.getDurationMicrosecondsNoStop();
+  // To be called only by calculateSingle, depends on preparations steps done there
```
tahini, line 87: how does it depend? when is this called? I'm under the impression, but I could be wrong, that there are 2 real code paths above and the 2 call it.

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -13,125 +13,146 @@ namespace TrRouting
...
+
+    return result;
+  }
+
+  std::unique_ptr<AllNodesResult> Calculator::calculateAllNodes(RouteParameters &parameters, bool resetAccessPaths, bool resetFilters) {
+    reset(parameters, resetAccessPaths, resetFilters);
+
+    std::unique_ptr<AllNodesResult> result;
+
+    if (departureTimeSeconds > -1 && parameters.isForwardCalculation())
```
tahini, line 117: Same comment for condition

connection_scan_algorithm/src/calculator.cpp
```diff
@@ -13,125 +13,146 @@ namespace TrRouting
...
+      calculationTime = algorithmCalculationTime.getDurationMicrosecondsNoStop();
+
+    }
+    else if (arrivalTimeSeconds > -1)
+    {
+      std::unordered_map<Node::uid_t, JourneyStep> reverseAccessJourneysSteps;
+
+      departureTimeSeconds = -1;
+      //TODO maybe we can do something different in that case, like a query flag
+      // we need to make all trips usable when not coming from forward result because reverse calculation, by default, checks for usableTrips
```
tahini, line 138: Is this a new comment? You're coming directly here, I think. Could the trips really be unusable? Would this make usable trips that are unusable in the scenario (if the scenario says not metro, the metro trips are unusable)

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -87,27 +87,28 @@ namespace TrRouting
             
             if ((**connection).canUnboard())
             {
-              // TODO probably extract the reverseJourneysSteps[nodeArrival.uuid] instead of doing lookup multiple the time
+              // Extract journeyStep once from map
+              const JourneyStep & reverseStepsAtArrival = reverseJourneysSteps.at(nodeArrival.uid);
```
tahini: reverseStep* singular

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -171,18 +164,18 @@ namespace TrRouting
                   )
                   {                    
                     if (
-                      initialDepartureTimeSeconds == -1
+                      departureTimeSeconds == -1
```
tahini, line 167: That was previously initialDepartureTimeSeconds, which was assigned at the beginning and said " set initial departure time so we can find the latest possible departure time with reverse calculation later and still know the initial waiting time". Is the departureTimeSeconds here equivalent? Is it set anywhere? Should it not be parameters, or use the isForwardCalculation() to check that it is a departureTime

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -171,18 +164,18 @@ namespace TrRouting
...
                       ||
                       (nodeDepartureInNodesAccessIte != nodesAccess.end() &&
-                       connectionDepartureTime - nodeDepartureInNodesAccessIte->second.time - connectionMinWaitingTimeSeconds >= initialDepartureTimeSeconds
+                       connectionDepartureTime - nodeDepartureInNodesAccessIte->second.time - connectionMinWaitingTimeSeconds >= departureTimeSeconds
```
tahini, line 170: Here too, since we use the value, the requested value is in the parameters, unless departureTimeSeconds is the same value (or is used here to support old code path from odtrips)

tahini (commented):
Looks good, but there's a few questions because... I am not a compiler ;-)

connection_scan_algorithm/include/calculator.hpp
greenscientist: It used internal fields, so yes.

connection_scan_algorithm/src/calculator.cpp
greenscientist: Yes, but there's a couple of lines in reset.cpp that need to be removed first.

connection_scan_algorithm/src/calculator.cpp
greenscientist: yes, exactly, it will

connection_scan_algorithm/src/calculator.cpp
greenscientist: So the functions depends on either the following code bloc, 
   for (auto && tripIte : transitData.getTrips()) {
￼        const Trip & trip = tripIte.second;		￼        const Trip & trip = tripIte.second;
￼        tripsQueryOverlay[trip.uid].usable = true;		￼        tripsQueryOverlay[trip.uid].usable = true;
￼      }		￼      }

Or the fact that the usable field was set in forwardCalculation

connection_scan_algorithm/src/calculator.cpp
greenscientist: same answer

connection_scan_algorithm/src/calculator.cpp
greenscientist: No, I added the comment previously, it's not not. 
When you come after forweardCalculation, some trips are unusable

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: It's not set anywhere else other than in reset.cpp and calculation.cpp, so they are equivalent. 
the isForwardCalculation will be a better check in the future.

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: We will be able to change departureTimeSeconds by a call to TimeOfQuery in RouteParameter (Forgot the exact name of the function)

greenscientist:
This is the bit of code that prohibit me from changing further the usage of departureTimeSeconds: https://github.com/chairemobilite/trRouting/blob/v2c/connection_scan_algorithm/src/resets.cpp#L44-L47

Once the odTrip call is cleaned up, and these line removed, we will be good.

greenscientist:
fixed the  reverseStep* singular

connection_scan_algorithm/src/calculator.cpp
tahini: Is there a code path where forwardCalculation was called before calculateAllNodes?

greenscientist: forwardCalculation is called inside calculateAllNodes. Not 100% sure I get the question right

tahini (approved):
sounds good!

connection_scan_algorithm/src/calculator.cpp
tahini: We are in calculateAllNodes, in a basic else and we haven't called anything yet... or am I missing someghint?

greenscientist: In the else, yes, we don't call forwardCalculation, but that's why we need to set all the trip as usable.

greenscientist: (They are unusable by default)

tahini: > (They are unusable by default)

Really? That's a useful information to comment on in the code!

greenscientist: Do we really need to add a comment since the code is like this: 
  class TripQueryData {
  public:

    TripQueryData()
      : usable(false),

greenscientist: (I hate this TripQueryData overlay, but I haven't found a better way to represent all this data so far. Maybe when i understand the code a bit better)

PR #217, opened 2022-12-05
result: Keep the Node object in the AccessibleNode result

tahini:
Instead of only the uuid, the whole Node object is kept, so the name and code can also be part of the returned values.

Also, format the accessible node percentage at the result generation level instead of in the calculation.

### DIFF
--- a/connection_scan_algorithm/src/result_to_v1.cpp
+++ b/connection_scan_algorithm/src/result_to_v1.cpp
@@ -201,7 +201,7 @@ namespace TrRouting
     json["nodes"] = nlohmann::json::array();
     for (auto &node : result.nodes) {
       nlohmann::json nodeJson;
-      nodeJson["id"]                     = boost::uuids::to_string(node.nodeUuid);
+      nodeJson["id"]                     = boost::uuids::to_string(node.node.uuid);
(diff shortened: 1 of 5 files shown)


tahini:
Preparation for the v2 implementation...

connection_scan_algorithm/src/result_to_v1.cpp
```diff
@@ -201,15 +201,16 @@ namespace TrRouting
...
+      nodeJson["id"]                     = boost::uuids::to_string(node.node.uuid);
       nodeJson["arrivalTime"]            = Toolbox::convertSecondsToFormattedTime(node.arrivalTime);
       nodeJson["arrivalTimeSeconds"]     = node.arrivalTime;
       nodeJson["totalTravelTimeSeconds"] = node.totalTravelTime;
       nodeJson["numberOfTransfers"]      = node.numberOfTransfers;
       json["nodes"].push_back(nodeJson);
     }
     json["numberOfReachableNodes"] = result.numberOfReachableNodes;
-    json["percentOfReachableNodes"] = result.percentOfReachableNodes;
+    // Get a number with 2 decimals.
```
greenscientist, line 212: I feel there is probably a function to do this properly, but that's for V1 only, so let's not bother too much

greenscientist (approved):
Yes, less TODO!

PR #218, opened 2022-12-05
v2: Add the `v2/odTrips` route for the batch odTrip calculations

tahini:
This route is for the od_trips=1 of v1, with only the required parameters. It takes a data source ID corresponding to the data source that contains the OD trip, along with the od trip query data. For now, only the periods and the sample_ratio are supported.

The response type is defined, though the charge profiles return values are not. There are a few TODOs in this route still to be determined.

### DIFF
--- a/docs/APIv2/API.yml
+++ b/docs/APIv2/API.yml
@@ -151,4 +151,57 @@ paths:
             application/json:
               schema:
                 $ref: 'accessibilityResponse.yml#/access_query_error'
+
+  /v2/odTrips:
+    get:
+      description: Calculate in batch all or a subset of the odTrips
+      parameters:
+      - in: query
+        name: data_source_id
+        schema:
+          type: string
+        required: true
+        description: ID of the data source that contains the od trips definitions
+      - in: query
+        name: periods
+        schema:
+          type: string
+          pattern: '^\d+,\d+(,\d+,\d+)*$'
+        required: false
+        description: Comma-separated list of period start and end times, in seconds since midnight. More than one period can be specified, in which case the format is 'start1,end1,start2,end2...'. If set, only the odTrips starting within those periods will be calculated.
+      - in: query
+        name: sample_ratio
+        schema:
+          type: number
+          minimum: 0
+          maximum: 1
+        required: false
+        description: Specify a ratio of odTrips to calculate. If set, a random number of odTrips corresponding to this ratio will be calculated. The odTrips qualify with the other criteria
+      - $ref: "parameters.yml#/scenarioParam"
+      - $ref: "parameters.yml#/minWaitingTimeParam"
+      - $ref: "parameters.yml#/maxAccessTravelTimeParam"
+      - $ref: "parameters.yml#/maxEgressTravelTimeParam"
+      - $ref: "parameters.yml#/maxTransferTravelTimeParam"
+      - $ref: "parameters.yml#/maxTravelTimeParam"
+      - $ref: "parameters.yml#/maxFirstWaitingTime"
+      responses:
+        '200':
+          description: Successful query, but there may be no node
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: 'commonResponse.yml#/data_error'
+                  - $ref: 'odTripsResponse.yml#/successResponse'
+                discriminator:
+                  propertyName: status
+                  mapping:
+                    success: 'odTripsResponse.yml#/successResponse'
+                    data_error: 'commonResponse.yml#/data_error'
+        '400':
+          description: Query parameters are invalid
+          content:
+            application/json:
+              schema:
+                $ref: 'odTripsResponse.yml#/od_trips_query_error'
     
--- a/docs/APIv2/odTripsResponse.yml
+++ b/docs/APIv2/odTripsResponse.yml
@@ -1,0 +1,107 @@
+od_trips_query_error: # 'query_error' is a value for the status (discriminator)
+  required:
+    - status
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [query_error]
+    errorCode:
+      type: string
+      enum:
+        - 'EMPTY_SCENARIO'
+        - 'EMPTY_DATA_SOURCE'
+        - 'MISSING_PARAM_SCENARIO'
+        - 'MISSING_PARAM_DATA_SOURCE'
+        - 'INVALID_PERIODS'
+        - 'INVALID_SAMPLE_RATIO'
+        - 'INVALID_NUMERICAL_DATA'
+        - 'PARAM_ERROR_UNKNOWN'
+
+successResponse:
+  required:
+    - status
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [success]
+    query:
+      $ref: '#/odTripsQueryResponse'
+    result:
+      $ref: '#/odTripsResultResponse'
+
+odTripsQueryResponse:
+  type: object
+  properties:
+    dataSourceId:
+      type: string
+      description: ID of the data source from which come the odTrips that were calculated.
+    sampleRatio:
+      type: number
+      description: The ratio of odTrips from the data source for which a route was calculated. This a number between 0 and 1 inclusively.
+
+odTripsResultResponse:
+  type: object
+  properties:
+    odTrips:
+      type: array
+      items:
+        $ref: '#/odTripResult'
+    # TODO v1 also returns line and paths profiles. We don't support it yet, but when we do, add the documentation here
+
+odTripResult:
+  type: object
+  properties:
+    tripData:
+      type: object
+      properties:
+        id:
+          type: string
+          description: ID of this OD trip
+        internalId:
+          type: string
+          description: Internal ID to identify this OD trip
+        originActivity:
+          type: string
+          # TODO This could probably be a enum. Also, isn't the activity an atribute of the trip and not of each end?
+          description: Activity done at the origin of the trip
+        destinationActivity:
+          type: string
+          description: Activity done at the destination of the trip
+        declaredMode:
+          type: string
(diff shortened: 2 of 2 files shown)


tahini:
I think the line/path charge profile should be a separate API route, like the single route vs summary, which are 2 outputs from the same calculations, but the callers of each have very different needs.
@kaligrafy can you confirm profil de charge and od trips detailed results are 2 different things?

docs/APIv2/API.yml
```diff
@@ -151,4 +151,66 @@ paths:
...
+        description: ID of the data source that contains the od trips definitions
+      - in: query
+        name: periods
+        schema:
+          type: array
+          items:
+            type: number
+          minItems: 2
+        required: false
+        description: List of period start times and period end times, in seconds since midnight. If set, only the odTrips starting within those periods will be calculated
```
greenscientist: So, we can have multiple pairs of start/end times?

docs/APIv2/API.yml
```diff
@@ -151,4 +151,66 @@ paths:
...
+      - in: query
+        name: sample_ratio
+        schema:
+          type: number
+          minimum: 0
+          maximum: 1
+        required: false
+        description: Specify a ratio of odTrips to calculate. If set, a random number of odTrips corresponding to this ratio will be calculated. The odTrips qualify with the other criteria
```
greenscientist, line 179: DO we want to pass a seed parameter there? (I saw some code for seed usage, but I think it might have been commented out and removed)

docs/APIv2/API.yml
tahini: given the parameters handling, yes. But do we want it in the API? @kaligrafy is that really a feature?

docs/APIv2/API.yml
tahini: random implies possible repro, so I guess yes. Whether we use it or not is another story. But there's quite a lot of possible parameters that are not here yet... I took those used in the call from Transition only. In doubt, we don't put them, it's easier to increment an API than to remove stuff.

greenscientist: Except for testing and we can accomplish testing in other way, we'd need to see if our scientist friends would need that. But it would be easy to add a new api entry to point for a specific use case. 
And worst case, if people want something stable, they can filter the odtrip list before sending it to trRouting.

tahini:
@greenscientist anything left to do here? This just mirrors the usage we have of this API in v1, just so we can get rid of v1

docs/APIv2/API.yml
greenscientist: We should clarify this field a little bit more if we want to keep it as is. 
Is it start1, end1, start2, end2 or start1, start2, end1, end2. (Is suspect the first, but should be clear)

docs/APIv2/API.yml
```diff
@@ -151,4 +151,66 @@ paths:
...
+      - $ref: "parameters.yml#/maxFirstWaitingTime"
+      responses:
+        '200':
+          description: Successful query, but there may be no node
+          content:
+            application/json:
+              schema:
+                oneOf:
+                  - $ref: 'commonResponse.yml#/data_error'
+                  - $ref: 'odTripsResponse.yml#/success'
```
greenscientist: Where is the odTripsResponse.yml?

tahini: It stayed on my computer... thanks!

docs/APIv2/odTripsResponse.yml
```diff
@@ -0,0 +1,106 @@
...
+      # TODO Better description
+      description: Maximum hourly demand for a segment
+    maxSegmentTotalDemand:
+      type: number
+      # TODO Better description
+      description: Maximum total demand for a segment
+    totalTravelTime:
+      type: number
+      description: Total travel time of all od trips calculated, in seconds
+    # TODO Document the lines and paths profile
```
greenscientist: Do we want to do it now?
What I'm particularly interested is what happen to the result when with_profile is off? Do the corresponding keys disappear from the results or they are just empty?

docs/APIv2/odTripsResponse.yml
```diff
@@ -0,0 +1,106 @@
...
+  type: object
+  properties:
+    odTrips:
+      type: array
+      items:
+        $ref: '#/odTripsResult'
+    maxSegmentHourlyDemand:
+      type: number
+      # TODO Better description
+      description: Maximum hourly demand for a segment
```
greenscientist: Agreed, it needs a better description, I struggle to figure out what it should represent.

docs/APIv2/odTripsResponse.yml
```diff
@@ -0,0 +1,106 @@
...
+    maxSegmentTotalDemand:
+      type: number
+      # TODO Better description
+      description: Maximum total demand for a segment
+    totalTravelTime:
+      type: number
+      description: Total travel time of all od trips calculated, in seconds
+    # TODO Document the lines and paths profile
+
+odTripsResult:
```
greenscientist: Should we drop the s in this type

This feel to be the result of one trip and not multiple ?

docs/APIv2/odTripsResponse.yml
```diff
@@ -0,0 +1,106 @@
...
+      type: string
+      description: Activity done at the destination of the trip
+    declaredMode:
+      type: string
+      # TODO Can't there be multiple modes per trip?
+      description: Mode declared for this trip
+    expansionFactor:
+      type: number
+      description: Expansion factor represented by the person doing this trip
+    onlyWalkingTravelTime:
```
greenscientist: I don't think that trRouting compute that???

greenscientist (commented):
Some starting comments

docs/APIv2/odTripsResponse.yml
tahini: > Do we want to do it now? What I'm particularly interested is what happen to the result when with_profile is off? Do the corresponding keys disappear from the results or they are just empty?

I don't know who "we" is, but _I_ certainly don't want to get into this now... unless you for me to... The keys do disappear from the result if no profile is requested

docs/APIv2/odTripsResponse.yml
tahini: @kaligrafy Do you have an idea here? Are those fields (maxSegmentHourlyDemand and maxSegmentTotalDemand) only for the profiles use case?

docs/APIv2/odTripsResponse.yml
tahini: It does not compute it, but it's part of the odTrip's data

docs/APIv2/odTripsResponse.yml
greenscientist: They seem to be currently calculated even when calculateProfile is false currently. (But rely on the profile calculation. 

So if we keep this, we can probably drop the with_profile param and always return the profile. (Profile might be a lot of data,maybe not.  But the descrption of with_profile should make it clear that the profile is still calculated, just not copied in the answer.

docs/APIv2/odTripsResponse.yml
greenscientist: We should at least specificy that this is a direct copy of the data from the odTrip file.

tahini: The new api has a specific field to contain the result for the odTrip, genuisly named `result`. Perhaps we could put all those odTrip specific fields that describe the odTrip under a field called, with as much insight as the other one, `odTrip`.

greenscientist: yeah, could be. Make a clear demarkation from original data and things actually calculated.

greenscientist: (We will so need a V3 on this specific call :) )

tahini:
Rebased on the other v2 PR.

docs/APIv2/odTripsResponse.yml
```diff
@@ -0,0 +1,95 @@
...
+
+successResponse:
+  required:
+    - status
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [success]
+    result: '#/odTripsResponse'
```
greenscientist: Are we missing a query field in this response?

docs/APIv2/API.yml
```diff
@@ -151,4 +151,64 @@ paths:
...
+      - in: query
+        name: sample_ratio
+        schema:
+          type: number
+          minimum: 0
+          maximum: 1
+        required: false
+        description: Specify a ratio of odTrips to calculate. If set, a random number of odTrips corresponding to this ratio will be calculated. The odTrips qualify with the other criteria
+      - in: query
+        # TODO Is this a different query than the one without profile? Or can one caller want both results for charge profile AND individual od trips?
```
greenscientist: Should we change the TODO to say that this is not implemented and that we don't have any response that match ?

docs/APIv2/odTripsResponse.yml
```diff
@@ -0,0 +1,107 @@
...
+          description: Declared departure time of this trip, in seconds since midnight
+        declaredArrivalTime:
+          type: number
+          description: Declared arrival time of this trip, in seconds since midnight
+      # TODO When we have a batch calculation endpoint, this data's structure may be revisited. Now we just match the v1 returned data
+      description: Data describing the odTrip. It comes from the base odTrip in the data source.
+    result:
+      $ref: '#/odTripRouteResponse'
+
+odTripRouteResponse:
```
greenscientist: odTripRoute*Result* maybe?  It's inside a Result object. 
Then then it contains a reouteResponse.   I'm confused :)

docs/APIv2/API.yml
tahini: Removed, we'll just add the option when we support it.

greenscientist (approved):
Ok, let's try to implement this and see if it works :)

PR #219, opened 2022-12-06
v2: Remove "Meters" suffix for distances in the route return types

tahini:
It is documented that the values are in meters. The "Seconds" suffix was already dropped from the time fields. For consistency, the Meters suffix is not necessary.

And do not repeat the success status and base parameters for each alternatives. This requires splitting the API return types for routes in smaller object types that can be reused in the alternatives.

### DIFF
--- a/connection_scan_algorithm/src/result_to_v2.cpp
+++ b/connection_scan_algorithm/src/result_to_v2.cpp
@@ -97,7 +97,7 @@ namespace TrRouting
     stepJson["nodeCoordinates"] = {step.node.point->longitude, step.node.point->latitude};
     stepJson["arrivalTime"] = step.arrivalTime;
     stepJson["inVehicleTime"] = step.inVehicleTime;
-    stepJson["inVehicleDistanceMeters"] = step.inVehicleDistanceMeters;
+    stepJson["inVehicleDistance"] = step.inVehicleDistanceMeters;
     response = stepJson;
   }
 
@@ -107,7 +107,7 @@ namespace TrRouting
     stepJson["action"] = "walking";
     stepJson["type"] = step.walkingType == walking_step_type::ACCESS ? "access" : step.walkingType == walking_step_type::EGRESS ? "egress" : "transfer";
     stepJson["travelTime"] = step.travelTime;
-    stepJson["distanceMeters"] = step.distanceMeters;
+    stepJson["distance"] = step.distanceMeters;
     stepJson["departureTime"] = step.departureTime;
     stepJson["arrivalTime"] = step.arrivalTime;
     if (step.walkingType != walking_step_type::EGRESS) {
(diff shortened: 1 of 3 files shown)


kaligrafy:
I strongly suggest to keep the suffixes everywhere! I cannot count the number of times there were errors in a report because the analyst/writer did not use the correct units, even when documented. Having the units in the field name is an error saving goodness!

kaligrafy (changes requested):
I will not approve this sorry! Please keep the units everywhere in every field name (seconds, minutes, meters, anything)

kaligrafy:
Just a sidenote here: we should never consider that the person reading the output file is the same that did the query in the beginning, so we cannot assume that the file reader did read the API.

tahini:
Well documented APIs typically don't keep the suffixes for units, or at least I never saw it (like osrm for instance). Some APIs are accompanied by their units, and it may be configurable (like miles vs km). If that forces consumers of the API to read the specification, all the best!

Another side note, this API is meant to be machine-readable and is not directly dumped to the user as is, at least not in Transition, where we will still format departure/arrival time and we will still provide the data in the proper units, etc. The API is the interchange format of information between the client and the server, that is all.

kaligrafy:
I understand your point about the API, but there will always be people reading these file and keeping them as is without parsing them in another tool... Except adding some characters to fiel names, what would be the disadvantage of keeping the suffixes?

greenscientist:
I will agree with @tahini here, the unit does not have it's place in the field name. First of all, meters is "the" standard unit, so it should be obvious. 
Also, the unit could be a thing that is negociated between the client and the server hense nullify the name of the field in that case. I don't foresee that it's something that we will do. 
In addition, it saves bandwith, when you start the have lot of request, it can start to be significant.

greenscientist:
> Just a sidenote here: we should never consider that the person reading the output file is the same that did the query in the beginning, so we cannot assume that the file reader did read the API.

This is not an output file, it's the result of a request between a client and a server. It it was meant to be a file, we would need some metadata to define it some how, but this is not what we are doing here.

kaligrafy:
I understand that theorically, this is just an output for another parser, but there are always people intercepting this data and using it for analysis. And it can create bad and big errors (and it has in the past). I would at least keep the official abbreviation (M, S, Km, KmH, etc.) as a suffix to reduce fiel name length.

kaligrafy:
Keeping the units on the whole chain removes one hole from the swiss cheese (see swiss cheese model)

greenscientist:
No, because then it opens the door to have multiple field for the same data with multiple units.

kaligrafy:
I am pretty sure an american one day will use this data and think the results are in feet or in miles instead of meters and km, and will make bad errors... 
I have seen people making analyses with errors of factors 1000 (they switched meters and km!). I have seen this over and over with students and/or clients.

kaligrafy (approved):
I'll let it pass, but in EVERY output from any of our apps using trRouting (like Transition), I would like to get the units in the field name.

greenscientist:
just for the history, I've posted this in the wrong PR yesterday: 
Valhalla (no units): https://valhalla.readthedocs.io/en/latest/api/turn-by-turn/api-reference/#outputs-of-a-route
OSRM (no units): http://project-osrm.org/docs/v5.5.1/api/#match-service
GraphHopper (no units): https://docs.graphhopper.com/#operation/getRoute/alternative_route.max_share_factor

kaligrafy:
here is a question/answers related to the subject of including units in field names:
[Naming conventions](https://stackoverflow.com/questions/445191/should-we-put-units-of-measurements-in-attribute-names)

kaligrafy:
[Google naming conventions ](https://cloud.google.com/apis/design/naming_convention) says to always include units in the API field name for integer, except standard timstamps:
Quantities
Quantities represented by an integer type must include the unit of measurement.
xxx_{bytes|width_pixels|meters}

greenscientist:
> here is a question/answers related to the subject of including units in field names: [Naming conventions](https://stackoverflow.com/questions/445191/should-we-put-units-of-measurements-in-attribute-names)

Fields name is different that API

greenscientist:
> [Google naming conventions ](https://cloud.google.com/apis/design/naming_convention) says to always include units in the API field name for integer, except standard timstamps: Quantities Quantities represented by an integer type must include the unit of measurement. xxx_{bytes|width_pixels|meters}

That's one standard, among others.

tahini:
What if we added a field at the root of the response: `{ units: { distance: meters, time: seconds } }` and all fields ending with distance or time follow those units? but then we need to change the description of those fields to say it matches the unit in the response... That would be even less clear imho.

greenscientist:
> What if we added a field at the root of the response: `{ units: { distance: meters, time: seconds } }` and all fields ending with distance or time follow those units? but then we need to change the description of those fields to say it matches the unit in the response... That would be even less clear imho.

That feels a bit complex for the moment.

connection_scan_algorithm/src/result_to_v2.cpp
```diff
@@ -40,8 +40,9 @@ namespace TrRouting
   private:
     nlohmann::json response;
     RouteParameters& params;
+    bool statusRequired;
```
greenscientist: I don't see a use case in the code for this.  Should probably be keep outside of the unit commit

greenscientist (changes requested):

connection_scan_algorithm/src/result_to_v2.cpp
tahini: Indeed there are 2 small-ish changes here that I put together in one commit, the suffix and the base query data switch of place. i can do 2 PRs if you prefer.

tahini: Previously, in alternatives response, the origin/destination, time and time type and status were repeated for every alternatives, while the topmost response just had 'alternatives'. Now those parameter fields are at the root of the response, and each alternative has only the details.

greenscientist: I would appreciate if they are at least separate commits. 
(I will access the unit part easily, I might have more questions on the status part)

tahini:
This is just the unit part now.

greenscientist (approved):

PR #221, opened 2022-12-08
v2: Do not repeat result status in each alternatives

tahini:
In the alternatives route response, the status only appears at the root of the response. The query parameters are also only at the root of the response instead of repeated in each alternative.

### DIFF
--- a/docs/APIv2/accessibilityResponse.yml
+++ b/docs/APIv2/accessibilityResponse.yml
@@ -17,7 +17,7 @@ access_query_error: # 'query_error' is a value for the status (discriminator)
         - 'INVALID_NUMERICAL_DATA'
         - 'PARAM_ERROR_UNKNOWN'
 
-baseRouteResponse:
+accessibilityQueryResponse:
   type: object
   properties:
     place:
@@ -42,36 +42,37 @@ baseRouteResponse:
 NoRoutingFound:
   required:
     - status
-  allOf:
-  - $ref: '#/baseRouteResponse'
-  - type: object
-    properties:
-      status:
-        type: string
-        enum: [no_routing_found]
-      reason:
-        type: string
-        enum:
-          # Generic reason, when not possible to specify more
-          - NO_ROUTING_FOUND,
-          # No accessible node around the place
-          - NO_ACCESS_AT_PLACE,
-          # There is no service at this place with the query parameters
-          - NO_SERVICE_AT_PLACE,
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [no_routing_found]
+    query:
+      $ref: '#/accessibilityQueryResponse'
+    reason:
+      type: string
+      enum:
+        # Generic reason, when not possible to specify more
+        - NO_ROUTING_FOUND,
+        # No accessible node around the place
+        - NO_ACCESS_AT_PLACE,
+        # There is no service at this place with the query parameters
+        - NO_SERVICE_AT_PLACE
 
-success:
+successResponse:
   required:
     - status
-  allOf:
-  - $ref: '#/baseRouteResponse'
-  - $ref: '#/accessibilityResponse'
-  - type: object
-    properties:
-      status:
-        type: string
-        enum: [success]
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [success]
+    query:
+      $ref: '#/accessibilityQueryResponse'
+    result:
+      $ref: '#/accessibilityResultResponse'
 
-accessibilityResponse:
+accessibilityResultResponse:
   type: object
   properties:
     nodes:
--- a/docs/APIv2/routeResponse.yml
+++ b/docs/APIv2/routeResponse.yml
@@ -1,46 +1,55 @@
 NoRoutingFound:
   required:
     - status
-  allOf:
-  - $ref: '#/baseRouteResponse'
-  - type: object
-    properties:
-      status:
-        type: string
-        enum: [no_routing_found]
-      reason:
-        type: string
-        enum:
-          # Generic reason, when not possible to specify more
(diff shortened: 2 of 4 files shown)


greenscientist (changes requested):
I don't have a great explanation for it, but I really don't like the use of a flag "status_required" in this implementation. 
Maybe it's just the name of the flag that does not fit well. 
Maybe we just need to drop the whole concept of the visitors (which we will be able to do)

greenscientist:
I think the things that's worrying the most, is that we should have a derived class for the AlternativeResult processing.

tahini:
> I don't have a great explanation for it, but I really don't like the use of a flag "status_required" in this implementation. Maybe it's just the name of the flag that does not fit well. Maybe we just need to drop the whole concept of the visitors (which we will be able to do)

Indeed, as I start thinking of the implementation of the other endoint's return values, I think the visitor, at least for the base response, was not such a great idea after all (for steps, it is ok). Each endpoint has a slightly different implementation (see the route vs summary) and some of the classes simply throw a "Does not make sense to be here" exception. Static functions specific for each endpoint would probably be sufficient. But that's what we have now...

docs/APIv2/routeResponse.yml
```diff
@@ -72,80 +72,85 @@ baseRouteResponse:
...
       items:
-        $ref: '#/routeResponse'
-    alternativesTotal:
-      type: number
-      description: Total number of alternatives that were calculated, including the failed ones
+        $ref: '#/tripStep'
+
+alternativesResponse:
+  allOf:
+    - $ref: '#/baseRouteResponse'
```
tahini: In the idea, like in the odTrips, to differentiate calculated responses from query data, should the baseRouteResponse be under a single field, eg. `query`, and all the rest would be calculated fields? In which case, should also the responses be under a clean `result` field?

eg:
```
{ status: 'success', query: { ... query params }, result: { alternatives: [...], alternativesTotal: } }

**Review by tahini — COMMENTED** [review]
Since I'm about to implement an actual client that makes use of that API, I'd like for it not to be too beta... so let's clarify once and almost for all!

**Comment by tahini** [comment]
@greenscientist If the implementation of the API's result bugs you, we could concentrate on clarifying the API and I can rewrite all results_to_v2 implementations later.

File: docs/APIv2/routeResponse.yml (hunk 1 above)
Comment (greenscientist) [reply docs/APIv2/routeResponse.yml]: I'm really wondering, why are we sending back the query again? 
(My main question is is that helpful? If it might be easier to differenciate queries when we send a lot, ok, sure, but will we do that?

Comment (tahini) [reply docs/APIv2/routeResponse.yml]: 1- It confirms that the query parameters are ok (like if you get an unexpected reverse origin lat/lon, you might realize you messed up the query) and what was calculated is really what was expected. 
2- And eventually, we will do a lot of queries, when we can and it's parallel.

**Comment by tahini** [comment]
As discussed, all API endpoints will have the base response format of: 
```
  status: list_of_statuses,
  query: reminder of query parameters
  result: If status is success, there should be something,
  reason: If status is not success, one of various possible reason
```

Each endpoint will document each of those fields.

**Comment by tahini** [comment]
All current endpoint definitions have been updated.

File: docs/APIv2/accessibilityResponse.yml [hunk 2]
```diff
@@ -42,34 +42,35 @@ baseRouteResponse:
...
-        enum: [success]
+  type: object
+  properties:
+    status:
+      type: string
+      enum: [success]
+    query:
+      $ref: '#/accessibilityQueryResponse'
+    result:
+      $ref: '#/accessibilityResponse'
```
greenscientist: Should we call that accessiblityResultResponse ? (To mark the different from the Query response)

docs/APIv2/routeResponse.yml
```diff
@@ -24,23 +25,25 @@ NoRoutingFound:
...
+  properties:
+    status:
+      type: string
+      enum: [success]
+    query: 
+      $ref: '#/routeQueryResponse'
+    result: 
+      $ref: '#/routeOrAlternativesResponse'
 
 routeOrAlternativesResponse:
```
greenscientist: I'm not a fan of this routeOrAlternative Response, can we split it up ? So it's clear that we are getting either a single route or a list of alternative?
I would be open to have 2 differents endpoints if they are not already splitted.

docs/APIv2/accessibilityResponse.yml
tahini: Sure! I guess the comment applies to all...

greenscientist: yeah, let's make it more uniform.

docs/APIv2/routeResponse.yml
tahini: thinking, thinking............ We could also do like osrm-backend's route endpoints and return an array of alternative in every case. Just with only 1 result if alternatives is false. That would make it simpler for everyone. As a client, if you're only interested in the first result, you just use this one. This would seemlessly make the odTrips route support both alternatives or not (see one comment in that commit)

greenscientist: Yeah, I think I would be happy with that. Will need a good description, but seem simple enough.

docs/APIv2/routeResponse.yml
```diff
@@ -2,12 +2,13 @@ NoRoutingFound:
   required:
     - status
   allOf:
-  - $ref: '#/baseRouteResponse'
```
tahini, line 5: no need for allOf here

greenscientist (approved):
Feels its in a good shape. 

We might want to add in a document a description of the general structure of each response (Status, Query, Result). but can be added later as we continue working on it.

PR #224, opened 2022-12-12
v2: Update the summary json response to match API

tahini:
The API now has a status/query/result structure. The summary is updated to match the API.

Also, instead of saving all agency and lines data in the LineSummary, the reference to the trip is stored.

### DIFF
--- a/connection_scan_algorithm/src/alternatives_routing.cpp
+++ b/connection_scan_algorithm/src/alternatives_routing.cpp
@@ -109,197 +109,186 @@ namespace TrRouting
     spdlog::debug("  alternativesMaxTravelTimeRatio: ", params.alternativesMaxTravelTimeRatio);
     spdlog::debug("calculating fastest alternative...");
   
-    std::unique_ptr<RoutingResult> result = calculateSingle(parameters);
+    std::unique_ptr<SingleCalculationResult> result = calculateSingle(parameters);
 
-    // Can technically be allNodes or SingleCalculation, so we check the type
-    if (result.get()->resType == result_type::SINGLE_CALCULATION)
+    SingleCalculationResult& routingResult = *result.get();
+    AlternativesResult alternatives = AlternativesResult();
+
+    alternatives.alternatives.push_back(std::move(result));
+
+    alternativeSequence++;
+    alternativesCalculatedCount++;
+
+    LineVisitor visitor = LineVisitor();
+
+    //departureTimeSeconds = routingResult.departureTimeSeconds + routingResult.firstWaitingTimeSeconds - params.minWaitingTimeSeconds;
+
+    maxTravelTime = params.alternativesMaxTravelTimeRatio * routingResult.totalTravelTime + (parameters.isForwardCalculation() ? routingResult.departureTime - parameters.getTimeOfTrip() : 0);
+    if (maxTravelTime < params.minAlternativeMaxTravelTimeSeconds)
     {
-      SingleCalculationResult& routingResult = dynamic_cast<SingleCalculationResult&>(*result.get());
-      AlternativesResult alternatives = AlternativesResult();
+      maxTravelTime = params.minAlternativeMaxTravelTimeSeconds;
+    }
+    else if (maxTravelTime > routingResult.totalTravelTime + params.alternativesMaxAddedTravelTimeSeconds)
+    {
+      maxTravelTime = routingResult.totalTravelTime + params.alternativesMaxAddedTravelTimeSeconds;
+    }
+    // TODO: We should not create a whole new object just to update maxTravelTime. This parameter should be in the calculation specific parameters, which do not exist yet
+    Point* origin = parameters.getOrigin();
+    Point* dest = parameters.getDestination();
+    RouteParameters alternativeParameters = RouteParameters(std::make_unique<Point>(origin->latitude, origin->longitude),
+      std::make_unique<Point>(dest->latitude, dest->longitude),
+      parameters.getScenario(),
+      parameters.getTimeOfTrip(),
+      parameters.getMinWaitingTimeSeconds(),
+      maxTravelTime,
+      parameters.getMaxAccessWalkingTravelTimeSeconds(),
+      parameters.getMaxEgressWalkingTravelTimeSeconds(),
+      parameters.getMaxTransferWalkingTravelTimeSeconds(),
+      parameters.getMaxFirstWaitingTimeSeconds(),
+      parameters.isWithAlternatives(),
+      parameters.isForwardCalculation());
 
-      alternatives.alternatives.push_back(std::move(result));
+    //params.departureTimeSeconds = departureTimeSeconds;
 
-      alternativeSequence++;
-      alternativesCalculatedCount++;
+    spdlog::debug("  fastestTravelTimeSeconds: {}", routingResult.totalTravelTime);
 
-      LineVisitor visitor = LineVisitor();
+    // Get the best result and extract the lines from it
+    // We will generate all the combinations of those lines and redo the calculation
+    // with each of the combination of line excluded
+    LineVector foundLines = routingResult.accept(visitor);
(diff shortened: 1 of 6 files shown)


connection_scan_algorithm/src/result_to_v2_summary.cpp
```diff
@@ -172,28 +153,36 @@ namespace TrRouting
...
     json["status"] = STATUS_SUCCESS;
-    json["origin"] = { params.getOrigin()->longitude, params.getOrigin()->latitude };
-    json["destination"] = { params.getDestination()->longitude, params.getDestination()->latitude };
-    json["timeOfTrip"] = params.getTimeOfTrip();
-    json["timeType"] = params.isForwardCalculation() ? 0 : 1;
+
+    // Query parameters for response
+    nlohmann::json queryJson;
+    queryJson["origin"] = { params.getOrigin()->longitude, params.getOrigin()->latitude };
+    queryJson["destination"] = { params.getDestination()->longitude, params.getDestination()->latitude };
```
greenscientist: We should create a function "toString" or better "toJson" for the Point object, so that we don't risk inverting them by error.

greenscientist (approved):

connection_scan_algorithm/src/result_to_v2_summary.cpp
tahini: They are inverted in v1, so this function should be in the v2 result files. 

By the way, what's the best way to organize the files/functions/classes for the result_to_v2 of all endpoints? 1 .hpp with all declarations, then a different .cpp for each endpoint? Or 1 .hpp + .cpp per endpoint (as it is currently with the summary and route endpoints)? That would mean + 1 .cpp for common functions like the Point toJson. And are static methods in a class a good approach for each endpoint, like currently?

greenscientist: Ok, yes, only in the V2 file for the Point to Json. That make it a bit less useful, I don't know if there are many location for that.

greenscientist: So for file organisation, normally: 1 hpp + 1 cpp per class (TrRouting is a bad example in general)

Normally, I would have said, one class per result type. But handling Json is a bit weird. I would need to think about the situation so juge if the current approach is ok or bad.

connection_scan_algorithm/src/result_to_v2_summary.cpp
```diff
@@ -157,12 +130,20 @@ namespace TrRouting
...
-    json["origin"] = { params.getOrigin()->longitude, params.getOrigin()->latitude };
-    json["destination"] = { params.getDestination()->longitude, params.getDestination()->latitude };
-    json["timeOfTrip"] = params.getTimeOfTrip();
-    json["timeType"] = params.isForwardCalculation() ? 0 : 1;
-    json["nbAlternativesCalculated"] = 0;
-    json["lines"] = nlohmann::json::array();
+
+    // Query parameters for response
+    nlohmann::json queryJson;
+    queryJson["origin"] = { params.getOrigin()->longitude, params.getOrigin()->latitude };
```
tahini: query response are same both for no_routing and success, they could be in a separate function. Same for the Point to json I guess, even if those functions are [for now] repeated for every endpoint.

connection_scan_algorithm/src/result_to_v2_summary.cpp
```diff
@@ -64,17 +42,22 @@ namespace TrRouting
   };
 
   /**
-   * @brief Visitor for the result object, returns a map of line uuid to LineSummary
+   * @brief Visitor for the result object, returns a map of line uuid to
+   * LineSummary
+   *
+   * TODO: When the alternatives are of concrete SingleCalculationResult instead
```
tahini: The types ARE concrete, even though still a ptr, so I think we can already remove the visitor pattern. I'll update...

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -109,197 +109,186 @@ namespace TrRouting
     spdlog::debug("  alternativesMaxTravelTimeRatio: ", params.alternativesMaxTravelTimeRatio);
     spdlog::debug("calculating fastest alternative...");
   
-    std::unique_ptr<RoutingResult> result = calculateSingle(parameters);
+    std::unique_ptr<SingleCalculationResult> result = calculateSingle(parameters);
 
-    // Can technically be allNodes or SingleCalculation, so we check the type
-    if (result.get()->resType == result_type::SINGLE_CALCULATION)
```
tahini, line 115: I removed 2 type checking ifs in this function, above and after alternative calculation, as well as change the return type of calculateSingle. The rest is just indentation changes.

connection_scan_algorithm/src/result_to_v2_summary.cpp
```diff
@@ -124,76 +58,126 @@ namespace TrRouting
...
 
-  void ResultToV2SummaryVisitor::visitSingleCalculationResult(const SingleCalculationResult& result)
+  /**
+   * @brief Parser for calculation results to add statistics to a line summary map
+   */
+  class SummaryResultParser {
+  private:
+    std::map<boost::uuids::uuid, LineSummary> lineSummaries;
+  public:
+    SummaryResultParser() {
```
greenscientist: We should pass the SingleCalculationResult directly to the constructor. That way, we don't have an object that will not be valid for some time. 
We can then set parseSingleCalculationResult as private (if we want. Depends if we like the object to be mutable or not.

connection_scan_algorithm/src/result_to_v2_summary.cpp
```diff
@@ -124,76 +58,126 @@ namespace TrRouting
...
         } else {
-          response.at(summary.lineUuid).count++;
+          lineSummaries.at(summary.trip.line.uuid).count++;
         }
       }
     }
   }
 
-  void ResultToV2SummaryVisitor::visitAlternativesResult(const AlternativesResult& result)
+  std::array<double, 2> pointToJson(Point * point)
```
greenscientist: const Point. 
Also I would prefer passing a reference instead of a pointer. 
If you want to keep the pointer, you will need to check that it's valid before the return

greenscientist (changes requested):
small changes

connection_scan_algorithm/src/result_to_v2_summary.cpp
tahini: There will be a singleton object of that class that parses multiple calculation results, one for each alternatives, so it will be the sum of all alternatives line. The object is valid from the start, it just contains an empty map, but since constructing it means there is either one single result or potentially many alternatives, it won't be the case...

connection_scan_algorithm/src/result_to_v2_summary.cpp
tahini: done. There's probably a legacy reason why origin and destination are pointers... I don't know if we can drop it, or wait for v1 to be dropped before making them a ref

connection_scan_algorithm/src/result_to_v2_summary.cpp
greenscientist: Ok, it's not really just a parser then.

greenscientist: Maybe SummaryResultGenerator??
Or LineAccumulator?
(ALso Singleton won't work as we will have multiple threads...)

tahini: LineAccumulator then. It's not a singleton in the pure sense. Just one single instance per result generation. It can be multi-threaded all right.

connection_scan_algorithm/src/result_to_v2_summary.cpp
greenscientist: Yeah, we can probably sweap around at some point and change that.

greenscientist (approved):

PR #225, opened 2022-12-12
Add index.html page for redoc page documenting the v2 API

### LINKED ISSUE #222: v2: Add a github pages for our APIs
When the v2 API is fixed, we can create a page to show the documentation, like a real API. See https://github.com/eclipse-cdt-cloud/trace-server-protocol/tree/gh-pages for an example of how to setup a github page using redoc.

tahini:
Fixes #222

This will create a formatted web page for the v2 API documentation.

### DIFF
--- a/docs/APIv2/index.html
+++ b/docs/APIv2/index.html
@@ -1,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>TrRouting API v2</title>
+    <!-- needed for adaptive design -->
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <!--
+    ReDoc doesn't change outer page styles
+    -->
+    <style>
+      body {
+        margin: 0;
+        padding: 0;
+      }
+    </style>
+  </head>
+  <body>
+    <redoc spec-url='https://raw.githubusercontent.com/chairemobilite/trRouting/v2c/docs/APIv2/API.yml'></redoc>
+    <script src="https://cdn.jsdelivr.net/npm/redoc/bundles/redoc.standalone.js"> </script>
+  </body>
+</html>
--- a/docs/index.html
+++ b/docs/index.html
@@ -1,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>TrRouting Documentation documentation</title>
+    <!-- needed for adaptive design -->
+    <meta charset="utf-8"/>
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <style>
+      body {
+        margin: 0;
+        padding: 0;
+      }
+    </style>
+  </head>
+  <body>
+    <a href="APIv2/index.html">API v2 documentation</a>
+  </body>
+</html>


index.html
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>TrRouting API v2</title>
```
greenscientist: I'm just confused, we have an index.html which seem generic, and we set the page title to say it's the V2 API? Should the title and this page be more generic?

index.html
```diff
@@ -0,0 +1,23 @@
...
+    -->
+    <style>
+      body {
+        margin: 0;
+        padding: 0;
+      }
+    </style>
+  </head>
+  <body>
+    <redoc spec-url='https://raw.githubusercontent.com/chairemobilite/trRouting/v2c/docs/APIv2/API.yml'></redoc>
```
tahini: It's not generic at all, it points to the v2 API here, so the title matches. But the index.html page needs to be in the root for the gh-pages branch to be able to create the page. Then, whenever the API is updated, we just need to merge the v2c branch in gh-pages for the update to be available on the project's github page.

greenscientist: Ok, all this need to be written down somewhere. Can be a commit in the .html or at least in the commit message.

greenscientist (changes requested):
Need more instructions.

index.html
greenscientist: We can link to some GH doc if we don't want everything.

greenscientist: This could also be automated with a GH action maybe.

greenscientist: btw, yes, it point to the V2 API, but the file name is index.html
When somebody will look at the file list, he will think that this is the root of a trRouting website.

tahini: We cannot rename this file. Would you prefer the index.html to point to some 1990-style page with "hello-trRouting" title and a link to the this page named apiv2.html? Somebody can 2022-ize the main page in future PR or add some trRouting page information.

I could also put this page directly in the gh-pages branch instead of adding it v2c. But then we need to maintain that branch separately. With the file in v2c, we can just have gh-pages follow v2c

tahini: > This could also be automated with a GH action maybe.

What do you mean by "This"? When the gh-pages branch is updated, github automatically updates the associated github page. We could automate the gh-pages branch update to always match v2c though.

greenscientist: Yes, that later part. The task of updating the gh-pages branch will be forgetten quickly. 
Eventually, if we have some automation there, it could also rename the file.

tahini: We can pick a theme, so no need for 1990-style if we want a pretty main page! https://pages.github.com/

greenscientist: > We cannot rename this file. Would you prefer the index.html to point to some 1990-style page with "hello-trRouting" title and a link to the this page named apiv2.html? Somebody can 2022-ize the main page in future PR or add some trRouting page information.
> 
> I could also put this page directly in the gh-pages branch instead of adding it v2c. But then we need to maintain that branch separately. With the file in v2c, we can just have gh-pages follow v2c

Yes, but we can do this later. (This feels like the OSRM page which is just a bit confusing.) 
I also have no idea what this looks like now. 

But having comments in the file that explain the situation would be a sufficient start if you don't want to do everything right now.

tahini:
This is what this looks like:

image

greenscientist:
Add what URL will that be?

tahini:
> Add what URL will that be?

file:///home/blabla/trRouting/docs/APIv2/index.html ! 
This will eventually be https://chairemobilite.github.io/trRouting/APIv2/index.html

greenscientist:
Just out of curiosity, it does not need to be at the root anymore?

tahini:
No! Actually, both the branch and the root are configurable after all! It can be either at the root or in a docs/ directory, like here. And iiuc, we can also tell to use the v2c branch, so there would be no need to sync anything. See https://github.com/chairemobilite/trRouting/settings/pages

greenscientist (approved):
sweet!

PR #227, opened 2022-12-15
csaTest: Remove the assertResultConversion from base class

tahini:
This function is implemented for each test and the result class of the parameter is not the same for each case. The `summary` test was already implementing an empty function for this one and adding its own local implementation. Each test class should just implement what it needs instead.


greenscientist (approved):
ok

PR #229, opened 2022-12-16
Update data source capnp fetcher to match Transition

tahini:
Also, change a log level for cache file and indendation of a test.


src/data_sources_cache_fetcher.cpp
```diff
@@ -99,6 +104,12 @@ namespace TrRouting
     } 
     catch (...) 
     {
+       std::exception_ptr eptr = std::current_exception(); // capture
```
greenscientist: What is this part of code playing with exception?
(Also alignment)

tahini: Sorry, it's just me trying to get a more meaningful error message, it was accidently committed...

greenscientist (approved):

PR #231, opened 2022-12-19
Implementation of the v2/accessibility

tahini:


tahini:
@kaligrafy Please, could you look at the last commit here to see if I'm saying anything silly: https://github.com/chairemobilite/trRouting/pull/231/commits/7fddf0acebe2365d0d15c5884522b500c38cda38

kaligrafy:
Perfect!

include/parameters.hpp
```diff
@@ -50,10 +54,8 @@ namespace TrRouting
       Type type;
   };
 
-  class RouteParameters {
+  class BaseParameters {
```
greenscientist: I think I would have called it CommonParameters instead, but it's not important enough to change. 
But can you add a comment with a short description to explain that this is for the common parts of parameters.

include/parameters.hpp
```diff
@@ -86,28 +88,21 @@ namespace TrRouting
...
       std::vector<std::reference_wrapper<const Mode>> exceptModes;
       std::vector<std::reference_wrapper<const Node>> exceptNodes;
-      bool withAlternatives; // calculate alternatives or not
       bool forwardCalculation; // forward calculation: default true. if false: reverse calculation, will ride connections backward (useful when setting the arrival time)
 
     public:
-      RouteParameters(std::unique_ptr<Point> orig,
-        std::unique_ptr<Point> dest,
-        const Scenario& scenario,
+      BaseParameters(const Scenario& scenario,
```
greenscientist: I think we could set the constructor as protected, so that we prohibit the instantion of only a baseparameter object.

connection_scan_algorithm/src/Makefile.am
```diff
@@ -10,11 +10,13 @@ libcsa_la_SOURCES = alternatives_routing.cpp \
 		   od_trips_routing.cpp \
 		   optimize_journey.cpp \
 		   parameters.cpp \
+		   parameters_base.cpp \
```
greenscientist: Please, name them the same way as the class. (route_parmeters.cpp)  Or rename the class to be ParametersRoute

greenscientist: If you have done this to ease the organisation, eventually all the parameter stuff can be into its own directory.

connection_scan_algorithm/src/parameters_accessibility.cpp
```diff
@@ -0,0 +1,193 @@
...
+          place = Point(std::stod(latitudeLongitudeVector[1]), std::stod(latitudeLongitudeVector[0]));
+        }
+        catch (...)
+        {
+          throw ParameterException(ParameterException::Type::INVALID_PLACE);
+        }
+      }
+
+      // times and date:
+      else if (parameterWithValue.first == "time_of_trip")
```
greenscientist: Isn't these thing common with Route?  They should probably be parsed in a common function

connection_scan_algorithm/include/constants.hpp
```diff
@@ -6,4 +6,14 @@
 const std::string STATUS_SUCCESS = "success";
 const std::string STATUS_NO_ROUTING_FOUND = "no_routing_found";
 
+// No routing reasons
```
greenscientist: Feels weird to have that in a generic contant.hpp file. Could we have a more specific place?

Could be a result_common.hpp file

greenscientist: Are they somewhat defined in a common file in the API.yml file? I want to ease the matching between the API definition and the code

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -48,14 +49,7 @@ namespace TrRouting
...
 
-    /**
-     * Prepare the data for the calculator and validate that there are at least
-     * some data to do calculations on
-     *
-     * @return A DataStatus corresponding to the data readiness. Only if the
-     * status is READY will queries be served on this data
-     */
-    void                    reset(RouteParameters &parameters, bool resetAccessPaths = true, bool resetFilters = true);
+    void reset(BaseParameters &parameters, std::optional<std::reference_wrapper<const Point>> origin, std::optional<std::reference_wrapper<const Point>> destination, bool resetAccessPaths = true, bool resetFilters = true);
```
greenscientist: Ouf. Hopefully we'll get rid of reset() soon...

greenscientist (changes requested):
A few things, mostly in the code organisation category

connection_scan_algorithm/include/constants.hpp
tahini: No, for the API yml file, each endpoint has its own enum of the no routing reason, they are not the same for each, there's a few differences (for the no service and no access). I didn't want to add another .hpp file, but why not!

connection_scan_algorithm/src/Makefile.am
tahini: Yes, I wanted all the parameters related files to be next to each other not to have to search for parameters test, but indeed another directory could work. Do I need an additional Makefile.am for this directory too or  just put the dir/file.ccp here?

greenscientist: For simplicity, you can put the dir/file cpp here. 

It would be a bit cleaner to create a Makefile.am there, but then we'd need to make and manage libraries.

connection_scan_algorithm/src/parameters_accessibility.cpp
tahini: I'll use aggregation instead of inheritance, it will allow to more easily parse the common parameters

connection_scan_algorithm/src/od_trips_routing.cpp
```diff
@@ -291,16 +291,8 @@ namespace TrRouting
...
-          parameters.getTimeOfTrip(),
-          parameters.getMinWaitingTimeSeconds(),
-          parameters.getMaxTotalTravelTimeSeconds(),
-          parameters.getMaxAccessWalkingTravelTimeSeconds(),
-          parameters.getMaxEgressWalkingTravelTimeSeconds(),
-          parameters.getMaxTransferWalkingTravelTimeSeconds(),
-          parameters.getMaxFirstWaitingTimeSeconds(),
           parameters.isWithAlternatives(),
-          parameters.isForwardCalculation());
+          parameters);
```
greenscientist, line 296: This parameters was confusing me a little bit. 
Maybe add a comment saying that we reuse the common part of the original request. 
(I was where this parameters come from, we don't have a CommonParameter around...)

include/parameters.hpp
```diff
@@ -176,6 +185,51 @@ namespace TrRouting
...
+       * If there are missing or invalid parameters, this function will throw a
+       * ParameterException error
+       **/
+      static AccessibilityParameters createAccessibilityParameter(std::vector<std::pair<std::string, std::string>> &parameters,
+                                                    const std::map<boost::uuids::uuid, Scenario> &scenarios
+      );
+  };
+
+  /**
+   * @brief Helper function to get an integer value from a string, or throw a
```
greenscientist: while you are here, could you add to the comment
from a string, using std::stoi, or throw...

I went around wondering why we don't use the std::stoi directly, but it seems we want a fonction that convert the exception.

connection_scan_algorithm/src/parameters/common_parameters.cpp
```diff
@@ -69,53 +62,7 @@ namespace TrRouting
...
-  RouteParameters::RouteParameters(const RouteParameters& routeParams):
-    CommonParameters(routeParams),
-    origin(std::make_unique<Point>(routeParams.origin.get()->latitude, routeParams.origin.get()->longitude)),
-    destination(std::make_unique<Point>(routeParams.destination.get()->latitude, routeParams.destination.get()->longitude)),
-    withAlternatives(routeParams.withAlternatives)
-  {
-  }
-
-  static int getIntegerValue(std::string strValue) {
+  int getIntegerValue(std::string strValue) {
```
greenscientist: I need to add myself a TODO to add the ability to have the key name in the exception thrown here.

include/parameters.hpp
```diff
@@ -176,6 +185,51 @@ namespace TrRouting
...
+   * INVALID_NUMERICAL_VALUE exception if the string is not an integer.
+   *
+   * @param strValue Integer string value
+   * @return int The converted integer
+   */
+  int getIntegerValue(std::string strValue);
```
greenscientist: Might worth it to add a TODO here to put this function into a base class for all parameter instead of having it floating around.

connection_scan_algorithm/include/result_to_v2_accessibility.hpp
```diff
@@ -0,0 +1,24 @@
+#ifndef TR_RESULT_TO_V2_ACCESSIBILITY_RESPONSE
+#define TR_RESULT_TO_V2_ACCESSIBILITY_RESPONSE
+
+#include "json.hpp"
+#include "routing_result.hpp"
```
greenscientist, line 5: Does not seem to be necessary in the .hpp, can be added simply to the necessary .cpp file.

greenscientist (approved):
Still a few things to changes, but that's mostly just comments and smallist thing, so accepting.

include/parameters.hpp
tahini: This is just the function declaration, we don't know that the implementation will be using std::stoi.

connection_scan_algorithm/include/result_to_v2_accessibility.hpp
tahini: It is necessary for the NoRoutingReason enum

include/parameters.hpp
greenscientist: The thing is without this information, there's no reason to have a getIntegerValue function. We should just use std::stol directly. So we need some explanation about why this function exist.

connection_scan_algorithm/include/result_to_v2_accessibility.hpp
greenscientist: ok my bad, confused routing_result and result_constant.

include/parameters.hpp
tahini: I added some comments, but the contract for the create*Parameters says to throw a ParameterException if the parameter values are missing or invalid. You can use std::stoi directly, but you have to handle and convert the exceptions yourself.

greenscientist: thanks

PR #235, opened 2023-01-09
Add brew upgrade to mac os build

greenscientist:

PR #236, opened 2023-01-10
Add a workaround to have github actions build work on MacOS

greenscientist:
There's a conflict in Homebrew with multiple python trying to install the same symlink. This workaround force install both python and it ignore the symlink error. (The error was for 2to3, so now important in our use case)


tahini:
Does this replace #235?

greenscientist:
> Does this replace #235?

Yes

tahini (approved):
If it works, looks good!

PR #238, opened 2023-02-17
performance: Cache connections by scenario

### LINKED ISSUE #237: Cache the connections per scenario
Here are some benchmarks: Same queries on the same scenario (no alternatives only)

* first one, from a cache that contains a lot of agencies and services, hence a lot of connections are disabled for the scenario
```
,Forward - no alternatives,Arrival time - no alternatives,Forward - alternatives,Arrival time - alternatives
Short Central Trip (Promenade Masson to Villeray),0.380708,0.472243961
Short Central Trip - outside time range,0.246711424,0.318548813
Long Central Trip (Ahuntsic to Sud-Ouest),0.519573410,0.596342278
Periphery of the transit network (Eastern bout de l'île),0.447866286,0.514036480
Crossing more than half the transit network territory - morning (From PAT to Lasalle),0.713987818,0.789459774
Crossing more than half the transit network territory - noon (From PAT to Lasalle)),0.752328518,0.809695648
Crossing more than half the transit network territory - evening (From PAT to Lasalle),0.822566123,0.986840491
Crossing more than half the transit network territory - night (From PAT to Lasalle),0.353444682,0.599654696
Crossing more than half the transit network territory - outside time range (From PAT to Lasalle),0.124488063,0.275306892
West to east of the transit network (Baie d'urfe to RDP),0.944703223
Central Trip - no alternative (Polytechnique to Villeray),0.362723524,0.475080659
Short Downtown trip,0.330942037,0.437289254
```

* Second one, with a cache that only contains the scenario to try
```
,Forward - no alternatives,Arrival time - no alternatives,Forward - alternatives,Arrival time - alternatives
Short Central Trip (Promenade Masson to Villeray),0.103170,0.134225507
Short Central Trip - outside time range,0.065073812,0.093628511
Long Central Trip (Ahuntsic to Sud-Ouest),0.136828939,0.179716258
Periphery of the transit network (Eastern bout de l'île),0.103071582,0.134082468
Crossing more than half the transit network territory - morning (From PAT to Lasalle),0.199409651,0.250991509
Crossing more than half the transit network territory - noon (From PAT to Lasalle)),0.198818508,0.250278157
Crossing more than half the transit network territory - evening (From PAT to Lasalle),0.215442965,0.320866866
Crossing more than half the transit network territory - night (From PAT to Lasalle),0.109408449,0.190605221
Crossing more than half the transit network territory - outside time range (From PAT to Lasalle),0.035504065,0.081881168
West to east of the transit network (Baie d'urfe to RDP),0.266705137
Central Trip - no alternative (Polytechnique to Villeray),0.090751771,0.131285010
Short Downtown trip,0.084808468,0.120763934
```

That's between 70% and 75% better. For a batch calculation that takes 12 hours, it falls down to 3.

Possibly, if each calculation only had the connections for a given scenario instead of the whole set of connection, we could get the performances of the second benchmark on data from the first.

tahini:
Fixes #237

This caches the connections for the queried scenario. Only the last queried scenario is kept to avoid having to manage cache eviction and memory usage.

At the beginning of the calculation, in the resetFilters step, the connection cache object is retrieved, and computed if a new scenario is requested. This object contains only the forward and reverse connections for this scenario. All others are ignored.

For instances with a lot of services this improves performance. For example, calculations are 75% faster for an instance that contained the complete Montreal area data for fall 2018 and 2019.

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -39,6 +39,7 @@ namespace TrRouting
   class AllNodesResult;
   class AlternativesResult;
   class TransitData;
+  class ConnectionSet;
   class Point;
 
   class Calculator {
@@ -116,6 +117,8 @@ namespace TrRouting
 
     std::unordered_map<Trip::uid_t, bool> tripsEnabled;
     std::unordered_map<Trip::uid_t, TripQueryData> tripsQueryOverlay; // Store addition trip info during a query processing
+    // A subset of the connections that are used for the current query.
+    std::shared_ptr<ConnectionSet> connectionSet;
 
     std::vector<NodeTimeDistance> accessFootpaths; // pair: accessNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters
     std::vector<NodeTimeDistance> egressFootpaths; // pair: egressNodeIndex, walkingTravelTimeSeconds, walkingDistanceMeters
--- a/connection_scan_algorithm/include/journey_step.hpp
+++ b/connection_scan_algorithm/include/journey_step.hpp
@@ -10,8 +10,8 @@ namespace TrRouting {
   //each of the step types. See issue #209
   class JourneyStep {
   public:
-    JourneyStep(std::optional<std::shared_ptr<Connection>> _finalEnterConnection,
-                std::optional<std::shared_ptr<Connection>> _finalExitConnection,
+    JourneyStep(std::optional<std::reference_wrapper<const Connection>> _finalEnterConnection,
+                std::optional<std::reference_wrapper<const Connection>> _finalExitConnection,
                 std::optional<std::reference_wrapper<const Trip>> _finalTrip,
                 int _transferTravelTime,
                 bool _sameNodeTransfer,
@@ -23,8 +23,8 @@ namespace TrRouting {
                                          transferDistance(_transferDistance) { }
 
     //TODO Why is there Final in the function name. Is that appropriate?
-    std::optional<std::shared_ptr<Connection>> getFinalEnterConnection() const {return finalEnterConnection;}
-    std::optional<std::shared_ptr<Connection>> getFinalExitConnection() const {return finalExitConnection;}
+    std::optional<std::reference_wrapper<const Connection>> getFinalEnterConnection() const {return finalEnterConnection;}
+    std::optional<std::reference_wrapper<const Connection>> getFinalExitConnection() const {return finalExitConnection;}
     std::optional<std::reference_wrapper<const Trip>> getFinalTrip() const {return finalTrip;}
     //TODO: Comment from @tahini.
     // So walking is tranferring in a transfer step, but walking is just walking in an access/egress step.
@@ -38,10 +38,10 @@ namespace TrRouting {
 
     //TODO Setting connections midway should also affected other parameters automatically
     // Their usage in the optimise function should be reviewed (Maybe in conjunction with #209)
-    void setFinalEnterConnection(std::shared_ptr<Connection> _connection) {
+    void setFinalEnterConnection(const Connection&  _connection) {
       finalEnterConnection = _connection;
     }
-    void setFinalExitConnection(std::shared_ptr<Connection> _connection) {
+    void setFinalExitConnection(const Connection& _connection) {
       finalExitConnection = _connection;
     }
 
@@ -61,8 +61,8 @@ namespace TrRouting {
     // If we have type of each step type, maybe this would be needed. #209
     bool hasConnections() const {return finalEnterConnection.has_value() && finalExitConnection.has_value();}
   private:
-    std::optional<std::shared_ptr<Connection>> finalEnterConnection;
-    std::optional<std::shared_ptr<Connection>> finalExitConnection;
+    std::optional<std::reference_wrapper<const Connection>> finalEnterConnection;
+    std::optional<std::reference_wrapper<const Connection>> finalExitConnection;
     std::optional<std::reference_wrapper<const Trip>> finalTrip;
     int transferTravelTime;
     bool sameNodeTransfer;
--- a/connection_scan_algorithm/src/forward_journey.cpp
+++ b/connection_scan_algorithm/src/forward_journey.cpp
@@ -31,9 +31,6 @@ namespace TrRouting
     if (forwardEgressJourneysSteps.count(resultingNode.uid) > 0) {
 
       std::deque<JourneyStep> journey;
-
-      std::shared_ptr<Connection> journeyStepEnterConnection;
-      std::shared_ptr<Connection> journeyStepExitConnection;
       std::vector<Leg>  legs;
 
       int totalInVehicleTime       { 0}; int transferArrivalTime    {-1}; int firstDepartureTime     {-1};
@@ -56,7 +53,7 @@ namespace TrRouting
       while (resultingNodeJourneyStep.hasConnections())
       {
         journey.push_front(resultingNodeJourneyStep);
-        bestAccessNode = resultingNodeJourneyStep.getFinalEnterConnection().value()->getDepartureNode();
+        bestAccessNode = resultingNodeJourneyStep.getFinalEnterConnection().value().get().getDepartureNode();
         resultingNodeJourneyStep = forwardJourneysSteps.at(bestAccessNode.value().get().uid);
       }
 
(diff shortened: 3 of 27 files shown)


tahini:
Friday evening code dump!

It appears to work, I added some unit tests, but haven't proof-read the code yet. Between const disqualifiers, shared_ptr and reference_wrappers, I'm just happy it finally compiles and works!

More tests are needed probably. @greenscientist  If you have a chance to get a quick look to let me know if there's anything fundamentally wrong here, let me know

kaligrafy:
What is the performance gain when switching scenario compared to before? I mean how much time it takes to filter the connections for a specific scenario + looping the connections vs only looping all connections?

tahini:
> What is the performance gain when switching scenario compared to before? I mean how much time it takes to filter the connections for a specific scenario + looping the connections vs only looping all connections?

Indeed, the first request for a scenario takes visibly longer, though subsequent requests are much faster. I'm not sure why, but there's probably something I can improve in this PR...

kaligrafy:
Could we create the filtered connections vector by scenario at the launch of the server so we don't need to filter later on? It would increase a bit the memory, but if you put pointers in the vectors, it should be small enough even for a large number of scenarios?
I think the main objective here is to make the requests faster, not save on RAM.

tahini:
Oddly enough, with small instances (where almost all connections are in the scenario), it takes longer with this patch than before (though I didn't try real benchmarks yet). I wasn't expecting any time gain or loss in that case (except on first run for scenario). But I'll run the real original benchmarks to have real data.

Also, about caching all scenarios at load, it doesn't scale, but trRouting doesn't scale anyway... I'll have to see how big the memory will increase if doing so with the eod complete instance. More benchmarking to do!

kaligrafy:
Or at least keep the previously used scenarios until a threshold or something, and then throw oldest filetered connections data.

tahini:
> Or at least keep the previously used scenarios until a threshold or something, and then throw oldest filetered connections data.

That’s the long term idea, but an lru cache is hard to implement and requires a lot of testing. This is to get quick results for our batch calculation users.

tahini:
For memory usage data, here's the memory usage of trRouting, while in action (ie with one cached scenario), with and without this PR. Sizes are in megabytes:

|    | v2c | This PR | overhead
|---|---|---|---
| benchmark data | 173 | 192 | 11%
| Small instance | 203 | 242 | 19%
| Large instance (eod) | 1946 | 1993 | 2.4%

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -47,7 +48,8 @@ namespace TrRouting
 
     std::map<std::string, int> benchmarking;
 
-    Calculator(const TransitData &_transitData);
+    // FIXME const was removed because calling getConnectionCacheForScenario apparently discards its const qualifier
```
greenscientist: We'll need to investigate that a bit more. 

Some time, with caching, using the mutable keyword is acceptable.

connection_scan_algorithm/include/journey_step.hpp
```diff
@@ -10,8 +10,8 @@ namespace TrRouting {
   //each of the step types. See issue #209
   class JourneyStep {
   public:
-    JourneyStep(std::optional<std::shared_ptr<Connection>> _finalEnterConnection,
-                std::optional<std::shared_ptr<Connection>> _finalExitConnection,
+    JourneyStep(std::optional<std::reference_wrapper<Connection>> _finalEnterConnection,
```
greenscientist: Would be interesting to do the switch from share_ptr to reference_wrapper as a separate commit.

connection_scan_algorithm/include/journey_step.hpp
```diff
@@ -38,10 +38,10 @@ namespace TrRouting {
 
     //TODO Setting connections midway should also affected other parameters automatically
     // Their usage in the optimise function should be reviewed (Maybe in conjunction with #209)
-    void setFinalEnterConnection(std::shared_ptr<Connection> _connection) {
+    void setFinalEnterConnection(Connection&  _connection) {
```
greenscientist: Not const?

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -81,27 +78,27 @@ namespace TrRouting
         if (journeyStep.hasConnections())
         {
           // journey tuple: final enter connection, final exit connection, final footpath
-          journeyStepEnterConnection = journeyStep.getFinalEnterConnection().value();
-          journeyStepExitConnection  = journeyStep.getFinalExitConnection().value();
-          const Node &journeyStepNodeDeparture   = journeyStepEnterConnection->getDepartureNode();
-          const Node &journeyStepNodeArrival     = journeyStepExitConnection->getArrivalNode();
+          Connection& journeyStepEnterConnection = journeyStep.getFinalEnterConnection().value().get();
```
greenscientist: I'm surprise that I had left them outside of if. You verified there was not another object with the same name that was masqued?

include/connection_cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+#include <boost/uuid/uuid.hpp>
+#include <memory>
+
+
+namespace TrRouting {
+
+class Connection;
+class Trip;
+
+class ConnectionCache {
```
greenscientist: You'll need at least a comment to describe the class

include/connection_cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+
+    // Contains iterator matching each hour of the day from the corresponding connections container.
+    // Used to speed up iterating the connections by skipping the connections that are too early or too late
+    std::vector<std::vector<std::reference_wrapper<Connection>>::const_iterator> forwardConnectionsBeginIteratorCache;
+    std::vector<std::vector<std::reference_wrapper<Connection>>::const_iterator> reverseConnectionsBeginIteratorCache;
+
+    void generateConnectionsIteratorCache();
+};
+
+class ScenarioConnectionCache {
```
greenscientist, line 23: How does this relate to ConnectionCache ? With a name like that I would have expected some inheritance

include/connection_cache.hpp
```diff
@@ -0,0 +1,63 @@
...
+       
+    };
+    virtual ~ScenarioConnectionCache() {}
+    
+    std::optional<std::shared_ptr<ConnectionCache>> get(boost::uuids::uuid uuid) const;
+    void set(boost::uuids::uuid uuid, std::shared_ptr<ConnectionCache> cache);
+
+  private:
+    std::optional<boost::uuids::uuid> lastUuid;
+    std::shared_ptr<ConnectionCache> lastConnection;
```
greenscientist: This tells me that ConnectionCache is not a cache, but a Connection "Collection". The class does not manage the caching mecanism

include/transit_data.hpp
```diff
@@ -67,6 +68,8 @@ namespace TrRouting {
     const std::map<boost::uuids::uuid, Scenario> & getScenarios() const {return scenarios;}
     const std::map<boost::uuids::uuid, Trip> & getTrips() const {return trips;}
 
+    std::shared_ptr<ConnectionCache> getConnectionCacheForScenario(const Scenario & scenario);
```
greenscientist: Yeah, this should be const, with some mutable cache object inside it.

greenscientist (changes requested):
We can fix the const correctness, no need to remove it. 

And yes, after reading the rest of the code, we need to rename to object ConnectionCache to something else. 
Also Implement the ScenarioCache into a separate .hpp/cpp

tahini:
@greenscientist consts are back, reference_wrapper of const objects are extracted, ConnectionCache is renamed to ConnectionSet, with comments for each class. And the class is unit tested.

tests/connection_scan_algorithm/csa_test_data_fetcher.hpp
```diff
@@ -58,6 +58,7 @@ class TestDataFetcher : public TrRouting::DataFetcher
     inline static const boost::uuids::uuid serviceUuid = uuidGenerator("11111111-2222-3333-4444-555555666666");
 
     inline static const boost::uuids::uuid scenarioUuid = uuidGenerator("11111111-2222-2345-789a-555555666666");
+    inline static const boost::uuids::uuid scenario2Uuid = uuidGenerator("11111111-2222-2345-789a-555555666667");
```
greenscientist, line 61: Details, but I think it would be nice to add somewhere an assert that scenarioUuid != scenario2Uuid. 

(Just to make sure we don't accidentaly make them the same)

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -116,6 +117,7 @@ namespace TrRouting
 
     std::unordered_map<Trip::uid_t, bool> tripsEnabled;
     std::unordered_map<Trip::uid_t, TripQueryData> tripsQueryOverlay; // Store addition trip info during a query processing
+    std::shared_ptr<ConnectionSet> connectionSet;
```
greenscientist, line 121: Can have a small comment about it

greenscientist: I would had more comments about the name of this variable, but it will need to be moved out of the Calculator object to make the calculator multithreaded, so we'll make the change later.

include/connection_cache.hpp
```diff
@@ -0,0 +1,74 @@
+#ifndef TR_CONNECTION_CACHE
```
greenscientist, line 1: Can you split this file? Each class their file.

src/connection_cache.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (hour < CONNECTION_ITERATOR_CACHE_BEGIN_HOUR) {
+      return reverseConnections.cend();
+    } else if (hour > CONNECTION_ITERATOR_CACHE_END_HOUR - 1) {
+      return reverseConnections.begin();
+    }
+
+    return reverseConnectionsBeginIteratorCache[hour];
+  };
+
+  // Create a cache with a begin iterator which match the connection closest to the specified hour
```
greenscientist: I don't see a red part in other file for this code, but I think it's a copy/paste, is that expected

greenscientist (changes requested):

tahini:
The comments are addressed. Today I'll investigate removing code from the old code path and what can be so long for the first scenario request. That can be done as part of this PR or in another one on top of it, as you prefer. But as is, it works and is standalone.

greenscientist:
If you want to do the removal of the old code as a separate PR then we will need comments in this one about the copy/paste, where the code came from and also a TODO about code removal. 
It's really hard to follow code moving like that in a git blame, so we need to leave bread crumbs

greenscientist (approved):
ok looks good

PR #240, opened 2023-02-20
Fix the cache being fetched twice

### LINKED ISSUE #239: Cache fetched twice?
I was just restarting trRouting while watching the logs, and it seems the cache is fetched twice. A first time, after the "Sorting connections..." message, it goes back to "Initializing modes..." and logs are repeated until the server is ready.

I'm starting trRouting from Transition, on the v2c branch.

tahini:
Fixes #239

The `loadAllData` function should be called only once in the TransitData constructor.

### DIFF
--- a/src/transit_data.cpp
+++ b/src/transit_data.cpp
@@ -20,9 +20,8 @@ namespace TrRouting {
   TransitData::TransitData(DataFetcher& fetcher) :
     dataFetcher(fetcher)
   {
-    
-    loadAllData();
-    if(loadAllData() != DataStatus::READY ) {
+    DataStatus loadStatus = loadAllData();
+    if (loadStatus != DataStatus::READY) {
       //TODO For now, don't throw on error because Transit server expect to get the dataStatus
       //throw std::exception("Incomplete transit data");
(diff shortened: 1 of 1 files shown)


kaligrafy (approved):

tahini:
@greenscientist  any idea what's the problem with macos builds? You temporarily fixed something recently. Is it related? though this one seems to be about pkg-config...

greenscientist:
About the macos build, maybe the pkg-config package was there by default in the image and not there anymore. I imagine we should try to add it manually in our build

src/transit_data.cpp
```diff
@@ -20,9 +20,8 @@ namespace TrRouting {
   TransitData::TransitData(DataFetcher& fetcher) :
     dataFetcher(fetcher)
   {
-    
-    loadAllData();
-    if(loadAllData() != DataStatus::READY ) {
+
+    if (loadAllData() != DataStatus::READY) {
```
greenscientist: Instead, can you put the result of loadAllData in a variable and check that in the if.

greenscientist (changes requested):

greenscientist (approved):

PR #241, opened 2023-02-20
Add pkg-config to homebrew package list

greenscientist:
It's not installed anymore by default, so add it manually


tahini (approved):
If tests pass of course...

PR #243, opened 2023-02-24
Cleanup and optimization after the scenario cache PR

tahini:
* Removed unused code
* Keep only disabled trips for each calculation instead of a boolean value for all trips all the time
* Let TransitData manage a vector of the Connection objects and forward and reverse connections can now use reference_wrapper to const Connection instead of shared_ptr.

Here are some benchmark results of this PR (in rough average, showing the performance gain), vs the preCache and the scenario cache PR, with various capnp cache data
|    | vs preCache, single | vs preCache, alternatives | vs scenario cache, single | vs scenario cache, alternatives
|----|----|----|----|----|
|benchmark cache| 16 to 50% | 15 to 50% | 20 to 50% | 15 to 45% |
| large cache (eod) | 65 to 95% | precache alternatives did not work | 17 to 30% | 15 to 30% |

For memory, if I take the data I mentioned in https://github.com/chairemobilite/trRouting/pull/238#issuecomment-1437458333
|    | preCache | PR #238 | This PR
|---|---|---|---
| benchmark data | 173 | 192 | 121 |
| Small instance | 203 | 242 | 140 | 
| Large instance (eod) | 1946 | 1993 | 1000 |

### DIFF
--- a/connection_scan_algorithm/include/calculator.hpp
+++ b/connection_scan_algorithm/include/calculator.hpp
@@ -115,7 +115,9 @@ namespace TrRouting
     std::unordered_map<Node::uid_t, NodeTimeDistance> nodesAccess; // travel time/distance from origin to accessible nodes
     std::unordered_map<Node::uid_t, NodeTimeDistance> nodesEgress; // travel time/distance to reach destination;
 
-    std::unordered_map<Trip::uid_t, bool> tripsEnabled;
+    // Field used at the time of the query, mostly for alternatives, which deactivate some of the scenario trips. Typically, there are less disabled trips and enabled ones, so we keep only those
+    std::unordered_map<Trip::uid_t, bool> tripsDisabled;
+    bool isTripDisabled(Trip::uid_t uid) const { return tripsDisabled.find(uid) != tripsDisabled.end(); }
     std::unordered_map<Trip::uid_t, TripQueryData> tripsQueryOverlay; // Store addition trip info during a query processing
     // A subset of the connections that are used for the current query.
     std::shared_ptr<ConnectionSet> connectionSet;
--- a/connection_scan_algorithm/src/resets.cpp
+++ b/connection_scan_algorithm/src/resets.cpp
@@ -258,6 +258,7 @@ namespace TrRouting
     // exclusions than the scenario (the combinations of lines). It is not
     // redundant with the one in the connection cache generator
     // TODO: This code is very similar to the one in getConnectionsForScenario, extract it
+    tripsDisabled.clear();
     for (auto & tripIte : connectionSet.get()->getTrips())
     {
       const Trip & trip = tripIte.get();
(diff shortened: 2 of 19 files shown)


tahini:
I'm honestly surprised, I wasn't expecting such performance and memory gain with what is here, but the tests pass and it works with Transition.

src/transit_data.cpp
```diff
@@ -134,80 +133,22 @@ namespace TrRouting {
...
 
     // Copy the connections to both forward and reverse vectors
     forwardConnections.clear();
     reverseConnections.clear();
     for (size_t i=0; i<connections.size(); i++)
     {
-      std::shared_ptr<Connection> reverseConnection = connections[i];
-      forwardConnections.push_back(std::move(connections[i]));
-      reverseConnections.push_back(std::move(reverseConnection));
+      forwardConnections.push_back(connections[i]);
```
tahini, line 150: Could this impact the memory usage? I don't know why there was a local copy of the shared_ptr before. Now that it's a reference, my guess is that it's ok to take the vector value directly, no?

tahini:
@greenscientist ping on this PR

greenscientist:
sorry, I've completely missed that.

connection_scan_algorithm/include/calculator.hpp
```diff
@@ -115,7 +115,9 @@ namespace TrRouting
     std::unordered_map<Node::uid_t, NodeTimeDistance> nodesAccess; // travel time/distance from origin to accessible nodes
     std::unordered_map<Node::uid_t, NodeTimeDistance> nodesEgress; // travel time/distance to reach destination;
 
-    std::unordered_map<Trip::uid_t, bool> tripsEnabled;
+    // Field used at the time of the query, mostly for alternatives, which deactivate some of the scenario trips. Typically, there are less disabled trips and enabled ones, so we keep only those
+    std::unordered_map<Trip::uid_t, bool> tripsDisabled;
+    bool isTripDisabled(Trip::uid_t uid) { return tripsDisabled.find(uid) != tripsDisabled.end(); }
```
greenscientist: This function can be const

connection_scan_algorithm/src/resets.cpp
```diff
@@ -258,6 +258,7 @@ namespace TrRouting
     // exclusions than the scenario (the combinations of lines). It is not
     // redundant with the one in the connection cache generator
     // TODO: This code is very similar to the one in getConnectionsForScenario, extract it
```
greenscientist, line 260: Is the TODO still true?

src/transit_data.cpp
greenscientist: Seems fine

greenscientist (commented):
Just a couple of minor things

connection_scan_algorithm/src/resets.cpp
tahini: Yeah, both codes still look like a copy paste of each other. Something can be extracted from it (for each trip in a set, check the only/exclude from something else)

greenscientist (approved):

PR #247, opened 2023-03-20
csa: Fix the alternatives' maximum travel time

### LINKED ISSUE #246: csa: Max travel time for alternatives may be larger than received parameters
In the alternativesRouting method, we calculate the alternatives' max travel time as a ratio of the travel time of the first calculation's returned time, but we don't max it at the value in the original parameters, causing the alternatives times to be larger than requested.

tahini:
Fixes #246

The maximum travel time for the alternatives calculation should be bounded by the value received in parameter and not just by a function of the first calculation.

Add a unit test that previously failed and passes with this fix.

### DIFF
--- a/connection_scan_algorithm/src/alternatives_routing.cpp
+++ b/connection_scan_algorithm/src/alternatives_routing.cpp
@@ -122,7 +122,7 @@ namespace TrRouting
     LineVisitor visitor = LineVisitor();
 
     //departureTimeSeconds = routingResult.departureTimeSeconds + routingResult.firstWaitingTimeSeconds - params.minWaitingTimeSeconds;
-
+    // TODO Extract the max travel time calculation to a function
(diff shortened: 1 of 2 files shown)


kaligrafy (approved):

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -132,6 +132,7 @@ namespace TrRouting
     {
       maxTravelTime = routingResult.totalTravelTime + params.alternativesMaxAddedTravelTimeSeconds;
     }
+    maxTravelTime = std::min(maxTravelTime, parameters.getMaxTotalTravelTimeSeconds());
```
greenscientist, line 135: this max travel time is complicated. Can we put it in a function?  
(At less add a TODO for that.

tahini: I'll go for the TODO, the params and parameters variables scare me...

Issue #248, opened 2023-08-04
review log levels of some statements


tahini:
When logging info, we have a lot of `calculating alternative x from a total of y ...`, but nothing else.

We should revisit our log statements to allow high level information in `info` level, then increase for log statements for `debug` and `trace`.

PR #249, opened 2023-08-07
Change log levels of endpoint entry/exit

tahini:
Part of #248

The alternatives calculation statements are set to 'debug' instead of 'info'.

The log statement at the beginning and end of the v2/* routes are set to 'info' and parameters are removed, spdlog will associate a timestamp with the log, so no need to display the application's time.

If a catchall exception occurs, the log level is 'error'.

TODO
Cleanup the 'debug' statements and set a few as 'trace'

### DIFF
--- a/connection_scan_algorithm/src/alternatives_routing.cpp
+++ b/connection_scan_algorithm/src/alternatives_routing.cpp
@@ -194,7 +194,7 @@ namespace TrRouting
           alternativeParameters.exceptLines.push_back(line);
         }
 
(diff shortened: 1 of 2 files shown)


kaligrafy (approved):

greenscientist:
@tahini , so you have a TODO in the commit message. Do you expect to push an update to this commit?

tahini:
> @tahini , so you have a TODO in the commit message. Do you expect to push an update to this commit?

Not in the short term, I'll open an issue

tahini:
But to completely fix #248, there's still the debug statements to look at

greenscientist (approved):

PR #251, opened 2023-08-23
csa: Make sure the connection arrival node is a candidate

### LINKED ISSUE #250: unknown map object in reverseJourneySteps
With the STLévis/RTC gtfs for the 2023 travel survey, it is not possible to calculate routes from Lévis to Quebec because an arrival node (station Beneva) is not present in the reverseJourneySteps map.

Any route from Lévis to Québec, with alternatives is affected. The fastest path is often found, but with the alternatives set to 'yes', the problem occurs.

Hypothesis: 
* would asymetric transferable node times make a node accessible in the forward direction, but not in the reverse?
* could reverse connections be unsorted so the arrival node is not yet in the list?

The cache causing the problem can be found here: https://github.com/chairemobilite/trRouting/releases/tag/untagged-17870e5ef2a95525d736

The following request causes the bug: http://localhost:4000/v2/route?origin=-71.26555890211999,46.71776856267837&destination=-71.32651370058815,46.77191936716429&scenario_id=832a1eb7-bdfd-46dc-9614-4e96fea417d4&time_of_trip=54000&time_type=0&alternatives=true&min_waiting_time=180&max_access_travel_time=900&max_egress_travel_time=900&max_transfer_travel_time=600&max_travel_time=10800

tahini:
fixes #250

Lookup the arrival node in the reverse journey steps map to make sure the arrival node of the connection is a candidate for the journey.

Otherwise, it throws an exception when calling the `at`, making the whole query fail.

### DIFF
--- a/connection_scan_algorithm/src/reverse_calculation.cpp
+++ b/connection_scan_algorithm/src/reverse_calculation.cpp
@@ -86,10 +86,12 @@ namespace TrRouting
           )
           {
             
-            if ((*connection).get().canUnboard())
+            // Make sure the arrival of the connection can unboard and is a candidate for the journey. Nodes are candidate if they can either reach the destination by foot, or a connection that can reach the destination within the specified parameters.
+            auto rjsIte = reverseJourneysSteps.find(nodeArrival.uid);
+            if ((*connection).get().canUnboard() && rjsIte != reverseJourneysSteps.end())
             {
               // Extract journeyStep once from map
-              const JourneyStep & reverseStepAtArrival = reverseJourneysSteps.at(nodeArrival.uid);
+              const JourneyStep & reverseStepAtArrival = rjsIte->second;
               if (!tripExitConnection.has_value()) // <= to make sure we get the same result as forward calculation, which uses >
               {
                 currentTripQueryOverlay.exitConnection = *connection;
--- a/connection_scan_algorithm/src/forward_calculation.cpp
+++ b/connection_scan_algorithm/src/forward_calculation.cpp
@@ -76,6 +76,7 @@ namespace TrRouting
             nodeDepartureTentativeTime = MAX_INT;
           }
 
+          // TODO Do we need to make sure the departure node exists in the forwardJourneySteps map? For the reverse calculation, we had to in order to fix issue https://github.com/chairemobilite/trRouting/issues/250 The issue may apply to forward too, but we have no example
           auto nodesAccessIte = nodesAccess.find(nodeDeparture.uid);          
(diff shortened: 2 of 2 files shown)


tahini:
This fixes the issues we have in the trRouting instances for the od surveys, but I think it may be a sign of a more general unequivalence in the rewrite... Maybe we should recheck where it says `//TODO Unclear if it's equivalent than previous code`. We used to have tuples with -1 everywhere by default, now we do lookups in maps... 

I'd love to be able to unit test this, but I have no idea how to reproduce for now. I can add an issue for it.

greenscientist:
> This fixes the issues we have in the trRouting instances for the od surveys, but I think it may be a sign of a more general unequivalence in the rewrite... Maybe we should recheck where it says `//TODO Unclear if it's equivalent than previous code`. We used to have tuples with -1 everywhere by default, now we do lookups in maps...
> 
> I'd love to be able to unit test this, but I have no idea how to reproduce for now. I can add an issue for it.

Yes, please add an issue!

greenscientist (commented):
Are we sure this is the right fix? 
Is the fact that the node not in the reverseJourneysSteps be the issue here instead?

tahini:
> Are we sure this is the right fix? Is the fact that the node not in the reverseJourneysSteps be the issue here instead?

I searched, bisected and wondered and tried to understand. Yes, I'm 80% sure this is the right fix, given the current code. Before, in the previous array version, all nodes were there, initialized to -1, so a check eventually discarded the node. In the new version, only candidate nodes are in the map, so we need to make sure it is present before retrieving it.

That being said, we could try to understand the algorithm better and see if it would be possible to rewrite in such a way that not all connections and nodes are considered in the first place, but I'm not sure it would work and it would take time.

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -86,10 +86,12 @@ namespace TrRouting
           )
           {
             
-            if ((*connection).get().canUnboard())
+            // Make sure the arrival of the connection can unboard and is a candidate for the journey
```
greenscientist: " is a candidate for the journey"
I'm not fully understanding this part of the comment. I think it needs a bit more explanation.

kaligrafy: Some stops are unboard only or board only, so we need to make sure the user can unboard at this stop (Pense à laval, les lignes STL, tu ne peux pas débarquer à Montréal et vice-versa)

tahini: Ah j'ai mis à jour le commit message, je vais essayer d'ajouter ici aussi...

kaligrafy (approved):

greenscientist (approved):

PR #253, opened 2023-08-29
Add an option to cache all connectionSets from all scenarios

### LINKED ISSUE #252: add a parameter for caching connections for all scenarios at launch
Needed for evolution validation for which we have week/saturday and sunday calls only, so we can cache them all three at launch.

greenscientist:
In a situation where you have multiple scenario and where you often switch between them, having only one connection set in the cache does not offer lot of perf improvement.

This add a --cacheAllConnectionSets parameter, which when set to true will cache all connection cache.

This is implemented with a new ScenarioConnectionCacheAll class. (the old one renamed to ScenarioConnectionCacheOne) The flag define which one the TransitData class will instanciate.

Issue #252

### DIFF
--- a/include/transit_data.hpp
+++ b/include/transit_data.hpp
@@ -52,7 +52,8 @@ namespace TrRouting {
 
   class TransitData {
   public:
-    TransitData(DataFetcher& dataFetcher);
+    TransitData(DataFetcher& dataFetcher, bool cacheAllScenarios = false);
+    virtual ~TransitData();
     DataStatus getDataStatus() const;
     
     // const data access functions
@@ -119,7 +120,8 @@ namespace TrRouting {
     std::vector<std::reference_wrapper<const Connection>> forwardConnections; // Forward connections, sorted by departure time ascending
     std::vector<std::reference_wrapper<const Connection>> reverseConnections; // Reverse connections, sorted by arrival time descending
 
-    mutable ScenarioConnectionCache scenarioConnectionCache = ScenarioConnectionCache();
+    //TODO Consider using a reference instead, making sure the object is always valid
+    mutable ScenarioConnectionCache *scenarioConnectionCache;
   };
 
 }
--- a/connection_scan_algorithm/include/program_options.hpp
+++ b/connection_scan_algorithm/include/program_options.hpp
@@ -16,6 +16,7 @@ namespace TrRouting
     std::string cachePath;
     int         port;
     bool        debug;
+    bool        cacheAllConnectionSets;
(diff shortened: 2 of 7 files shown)


greenscientist:
Tested with orsm by setting  bool OsrmFetcher::birdDistanceAccessibilityEnabled to true

Ran with ./src/trRouting --cachePath transition_od_mtl_2023/ --cacheAllConnectionSets=true --debug=1

time wget "http://localhost:4000/v2/route?origin=-73.603835,45.527467&destination=-73.580651,45.477921&scenario_id=68c38af2-91af-49da-8c49-7082f63ea127&time_of_trip=54000&time_type=0&alternatives=false&min_waiting_time=180&max_access_travel_time=900&max_egress_travel_time=900&max_transfer_travel_time=600&max_travel_time=10800"
time wget "http://localhost:4000/v2/route?origin=-73.603835,45.527467&destination=-73.580651,45.477921&scenario_id=f2ddf899-cea2-4ae7-9e60-f213cf86ce87&time_of_trip=54000&time_type=0&alternatives=false&min_waiting_time=180&max_access_travel_time=900&max_egress_travel_time=900&max_transfer_travel_time=600&max_travel_time=10800"

The first request for each scenario are about 
real	0m0,360s
user	0m0,004s
sys	0m0,009s

And the next ones are
real	0m0,027s
user	0m0,006s
sys	0m0,001s

kaligrafy (approved):

include/transit_data.hpp
```diff
@@ -119,7 +120,7 @@ namespace TrRouting {
     std::vector<std::reference_wrapper<const Connection>> forwardConnections; // Forward connections, sorted by departure time ascending
     std::vector<std::reference_wrapper<const Connection>> reverseConnections; // Reverse connections, sorted by arrival time descending
 
-    mutable ScenarioConnectionCache scenarioConnectionCache = ScenarioConnectionCache();
+    mutable ScenarioConnectionCache *scenarioConnectionCache;
```
tahini, line 124: It has to be a pointer, not a reference?

tahini (approved):
Thanks!

include/transit_data.hpp
greenscientist: It was easier, I was not sure I could initialize a reference that way. I'll check.

greenscientist: Still looking into how to do the init as a reference.

greenscientist: Added a TODO,

PR #256, opened 2023-08-31
Refactor the OSRM/BirdDistance management

### LINKED ISSUE #254: Allow to specify to use bird distance if osrm not present
Currently, when running from the command line, a running osrm instance is required. There's a config option to use bird distance, used in tests, but which can't be set from the command line.

greenscientist:
This: 
* Split The OsrmFetcher into 2 classes and a base class. The OSRM side and Bird distance (now Euclidean) are in their own class. We pass the desired object to the Calculator constructor. 
* Add a CLI option so active the Euclidean (Bird distance) mode
* Add a check for errors from OSRM. (Like if the server is not running)


include/euclideangeofilter.hpp
```diff
@@ -0,0 +1,23 @@
...
+    
+    virtual std::vector<NodeTimeDistance> getAccessibleNodesFootpathsFromPoint(const Point &point,
+                                                                       const std::map<boost::uuids::uuid, Node> &nodes,
+                                                                       int maxWalkingTravelTime,
+                                                                       float walkingSpeedMetersPerSecond,
+                                                                       bool reversed = false);    
+  };
+}
+
+#endif // TR_OSRM_GEO_FILTER
```
tahini: TR_EUCLIDEAN... fwiw

include/geofilter.hpp
```diff
@@ -0,0 +1,34 @@
...
+    virtual std::vector<NodeTimeDistance> getAccessibleNodesFootpathsFromPoint(const Point &point,
+                                                                       const std::map<boost::uuids::uuid, Node> &nodes,
+                                                                       int maxWalkingTravelTime,
+                                                                       float walkingSpeedMetersPerSecond,
+                                                                       bool reversed = false) = 0;
+  protected:
+    // Common utility functions
+    static std::tuple<float, float> calculateLengthOfOneDegree(const Point &point);
+    static float calculateMaxDistanceSquared(int maxWalkingTravelTime, float walkingSpeed);
+    //TODO WHy one point * and other &
```
tahini, line 28: Indeed good question, why? It was like that before?

src/osrmgeofilter.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+    auto lengthOfOneDegree = calculateLengthOfOneDegree(point);
+    float maxDistanceMetersSquared = calculateMaxDistanceSquared(maxWalkingTravelTime, walkingSpeedMetersPerSecond);
+    float distanceMetersSquared;
+
+    spdlog::debug("osrm with host {} and port {}", host, port);
+
+    std::string queryString = "/table/v1/" + mode + "/" + std::to_string(point.longitude) + "," + std::to_string(point.latitude);
+
+    for (auto &&[uuid,node] : nodes)
```
tahini, line 37: Could this part of the code actually use the euclidian distance? Instead of having to use the protected methods here to calculate something that really looks like the euclidian distance (I haven't compare the exact code).

tahini (commented):
Some questions and comments, otherwise it looks good. Did you test with/without transition?

include/geofilter.hpp
greenscientist: Yeah, that was just a copy paste from the previous code.

src/osrmgeofilter.cpp
greenscientist: hum, I guess you are right, need to look at the code more, but it seems we are doing the euclidian first to filter the nodes and then filter those with OSRM. 

(it hurts to know that the data is also in PostGIS usually, since that query would be so simple to do there. )

greenscientist: Ok, re-read the code in more detail. 
It does calculate the euclidian distance, but it does something different with the data. Using the common base functions seems the best way to go here. I'll try to add a comment

greenscientist: @tahini Let me know if the comment I added explain the difference sufficiently.

include/geofilter.hpp
```diff
@@ -0,0 +1,34 @@
...
+  public:
+    virtual std::vector<NodeTimeDistance> getAccessibleNodesFootpathsFromPoint(const Point &point,
+                                                                       const std::map<boost::uuids::uuid, Node> &nodes,
+                                                                       int maxWalkingTravelTime,
+                                                                       float walkingSpeedMetersPerSecond,
+                                                                       bool reversed = false) = 0;
+  protected:
+    // Common utility functions
+    static std::tuple<float, float> calculateLengthOfOneDegree(const Point &point);
+    static float calculateMaxDistanceSquared(int maxWalkingTravelTime, float walkingSpeed);
```
tahini, line 27: Eventually todo, don't refer to walking, it's just a travel time and speed

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -132,8 +126,18 @@ int main(int argc, char** argv) {
   //TODO We wanted to handle error in the constructor, but later part of this code expect a dataStatus
   // leaving as a todo
   DataStatus dataStatus = transitData.getDataStatus();
-  
-  Calculator calculator(transitData);
+
+  // Selection which geofilter to use. OSRM is the default one. Euclidean mostly used for debugging and testing
+  GeoFilter *geoFilter = 0;
```
tahini, line 131: This pointer never gets deleted. I guess it doesn't matter since it's constructed once in the main?

tahini (approved):
Some comments, but nothing to do about them

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: Yeah, it should be cleaned, but it does not really matter as you said since it lives for the whole duration of the program.

PR #265, opened 2023-10-24
Move nodesReverseTentativeTime to a vector

greenscientist:
Apply the same optimisation than nodesTentativeTime

### DIFF
--- a/connection_scan_algorithm/src/reverse_calculation.cpp
+++ b/connection_scan_algorithm/src/reverse_calculation.cpp
@@ -70,13 +70,7 @@ namespace TrRouting
           const Node &nodeArrival = (*connection).get().getArrivalNode();
 
           // Extract node arrival time
-          int nodeArrivalTentativeTime  = -1;
-          auto ite = nodesReverseTentativeTime.find(nodeArrival.uid);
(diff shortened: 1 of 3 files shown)


greenscientist:
We see an extra 15%-20% speed up for the summary call in the "forward" case and a good 50% speed up for the "reverse" case. 
Confirmed same results in my test cases

connection_scan_algorithm/src/reverse_calculation.cpp
```diff
@@ -147,8 +139,7 @@ namespace TrRouting
 
                 if (footpathTravelTime <= parameters.getMaxTransferWalkingTravelTimeSeconds())
                 {                  
-                  if ((trite == nodesReverseTentativeTime.end()) || //TODO Unclear if it's equivalent than previous code
```
tahini, line 150: Is this now equivalent to previous code? ;-)

tahini (approved):
Thanks

connection_scan_algorithm/src/reverse_calculation.cpp
greenscientist: Yeah, I went back in history and put back the structure we had before :)

PR #268, opened 2023-11-07
Don't copie the scenario filter into the commonparameters

greenscientist:
When creating the initial ConnectionSet, we fetch the filters directly from the scenario object. With the current code copying the filter data, we would reapply the same filters to the trips list in the reset functions, spending time for no use.


greenscientist:
Improvement is small, but still make a difference. 

Confirmed we have the same results

tahini (approved):

PR #269, opened 2023-11-07
Line UID

greenscientist:
Two commit here: 
* One small variable renaming to avoid confusing
* Addition of a local UID to the Line object, for about a 5% improvement

### DIFF
--- a/include/line.hpp
+++ b/include/line.hpp
@@ -14,6 +14,9 @@ namespace TrRouting
   class Line {
   
   public:
+
+    typedef int uid_t; //Type for a local temporary ID
+
     Line(const boost::uuids::uuid &auuid,
          const Agency &aagency,
          const Mode &amode,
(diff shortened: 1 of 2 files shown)


include/line.hpp
```diff
@@ -27,7 +30,8 @@ namespace TrRouting
       shortname(ashortname),
       longname(alongname),
       internalId(ainternalId),
-      allowSameLineTransfers(aallowSameLineTransfers) {}
+      allowSameLineTransfers(aallowSameLineTransfers),
+      uid(++global_uid) {}
```
tahini, line 34: global_uid++? to assign the 0 value or not?

tahini (commented):
just a question

include/line.hpp
greenscientist: Yeah, we skip the zero everywhere. Maybe we should create an issue to change it everywhere. 
(I should maybe juste make it a base class for all data object)

tahini: It used to be an ID for the object, it made sense to start at one, but now we use it as array index, it would make sense to start at 0... What is the future of this?

tahini (approved):

include/line.hpp
greenscientist: It's still an ID that points to a location in the array. We don't use push_back to add to the array, we use the assignment at a specific position. So the source of truth for the location is that ID.

greenscientist: (It's messy and should be encapsulated into some kind of object probably)

PR #273, opened 2023-11-16
Stack of commits that remove the V1 API

greenscientist:
Transition does not use the V1 API anymore, so we can do some clean up. 

This series of commit : 
* Remove the http handler for V1
* Clean the Parameters class usage
  * Some default parameters were moved to the CommonParameter class
  * An OdTripLegacyParameter was created with the left over
* The V1 specific unit test removed

We don't have a global params object anymore.


tahini (approved):
Nice red diff!

Looks good to me. Have you tested it outside the unit tests, especially the alternatives, which have changed a bit?

greenscientist:
Yes, I've tested with my test script with does route and summary calculation with alternatives and the results stay the same as before.

PR #275, opened 2023-11-20
Remove unused benchmarking object

greenscientist:


tests/benchmark_csa/benchmark_CSA_test.cpp
```diff
@@ -109,7 +109,6 @@ class BenchmarkCSATests : public ConstantBenchmarkCSATests
   {
     // TODO Shouldn't have to do this, a query is not a benchmark
     calculator->algorithmCalculationTime.start();
-    calculator->benchmarking.clear();
```
tahini, line 112: Should the TODOs be updated too?
Also, we keep the algorithmCalculation ? Is it still used for some purpose?
Maybe add some note in the commit message about this, as the `algorithmCalculationTime` seems to be used to assign to benchmarking.

tahini (commented):
See if TODOs still apply

tests/benchmark_csa/benchmark_CSA_test.cpp
greenscientist: I did remove them in the next PR

tahini (approved):
ok then

PR #276, opened 2023-11-21
Enable multithreaded calculation

### LINKED ISSUE #58: CSA's Calculator parameters are not thread safe
The calculator class has a params field which contains the parameters of the calculation (origin, destination, etc). This means only one calculation can be done at one time. Parallelizing the calculation will require those parameters to be outside the calculator class. The calculator class should contain only immutable data.

greenscientist:
Two commits: 
* First move the calculator object locally to each requests.
* Add an command line option to handle request via multiple threads

### DIFF
--- a/connection_scan_algorithm/src/transit_routing_http_server.cpp
+++ b/connection_scan_algorithm/src/transit_routing_http_server.cpp
@@ -136,9 +136,6 @@ int main(int argc, char** argv) {
     spdlog::info("Using OSRM for access/egress node time/distance");
   }
 
-  Calculator calculator(transitData, *geoFilter);
-  //TODO, should this be in the constructor?
-  calculator.initializeCalculationData();
   spdlog::info("preparing server...");
 
   //HTTP-server using 1 thread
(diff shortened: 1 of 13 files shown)


greenscientist:
This is on top of https://github.com/chairemobilite/trRouting/pull/275

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -340,11 +308,7 @@ int main(int argc, char** argv) {
       *serverResponse << "HTTP/1.1 200 OK\r\nAccess-Control-Allow-Origin: *\r\nContent-Type: application/json; charset=utf-8\r\nContent-Length: " << response.length() << "\r\n\r\n" << response;
       return;
     }
-
-    // prepare benchmarking and timer:
-    // TODO Shouldn't have to do this, a query is not a benchmark
-    calculator.algorithmCalculationTime.start();
-    calculator.benchmarking.clear();
+    Calculator calculator(transitData, *geoFilter);
```
tahini, line 311: So iiuc, there will be one Calculator object per request?

tahini (commented):
So... this is all it takes to make trRouting multi-threaded? (this and a year of PRs before :p )

One question though: The TransitData object has a mutable ConnectionCache object. Can't there be race conditions on this object or its content? What if a query comes and changes the cache during another thread's calculation?

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: exactly

greenscientist:
> One question though: The TransitData object has a mutable ConnectionCache object. Can't there be race conditions on this object or its content? What if a query comes and changes the cache during another thread's calculation?

Yeah, that's a good point.. I'll have to verify that part.

greenscientist:
So a couple of cases here: 
If we use the --cacheAllConnectionSets=true, then we don't have any issue. The cache for each scenario is actually constant and we don't really have a race here. (We could have 2 threads that compute the cache initially, but they will just store the exact same data twice. Not efficient, would be better with a lock, but still work ok. 
For the case where we only cache the last used connection set, I think it will still work correctly. When we get the shared pointer in the cache, we check that it's the right UUID when we get it. so if thte data cached is the right one, we will get the right data. At the end, we return the data that we just computed, so if the data was not in the cache, we will receive the correct data, even if another computed a different cache for a different UUID. Again, spreading a couple of mutex would be a good improvement just to be in the extreme safe case. 
I think we are good to deploy this code as is. Will definitely followup shortly which some lock improvements. 
What do you think?

greenscientist:
Added locks in the ScenarioCacheConnection objects

tahini (commented):
Looks good then.
Last question: did you run the [single-threaded] benchmarks to make sure there is no unexpected side-effects for the single-thread scenario?

greenscientist:
I'm quite sure I did run it, but I'm doing a run right now to make sure.

greenscientist:
@tahini, yup, results looks good with the benchmarks

tahini (approved):
Then you have my blessing! Thanks! Yé, multi-threading!!

Issue #75, opened 2021-07-26
Replace math.h include with cmath


greenscientist:
include/parameters.hpp wrongly use math.h instead of cmath in its include list
(Also that include should probably be in the cpp file instead of the hpp once

PR #278, opened 2023-11-25
Use cmath instead of math.h

greenscientist:


greenscientist:
This for #75 

We use our local MAX_INT in many places, so we our shorthand there.

tahini:
@greenscientist The issue said it should be included in the cpp file instead of the hpp. Is this still valid?

greenscientist:
Not in the current incarnation. The TrRouting::MAX_INT is used in so many places that it makes sense to keep it there. 

I've created https://github.com/chairemobilite/trRouting/issues/280 as a follow up to further clean up the MAX_INT usage.

tahini (approved):

Issue #74, opened 2021-07-26
Use an external lib for nlohmann json


greenscientist:
We could use this lib as a dependency instead of copying in it fully. 
Might make it easier to keep up to date.

kaligrafy:
Yes, i approve!

PR #279, opened 2023-11-25
Use external nlohmann-json library instead of copying the files in th…

greenscientist:
…e project

Distro have packages for the json library that we use. Instead of having to maintain a local file in the repo, we will rely on external files on the system.

We changed the #include reference to point the the standard files and added a check in the configure to make sure the file is available.
Also updated the github actions to install the right packages and updated the README to add the package to the install instructions


greenscientist:
For #74

kaligrafy (approved):

tahini:
Thanks! looks good, but what happens with Docker?

greenscientist:
@tahini , I think it was a sync issue. I had a few failure, but the latest build passed correctly when you check docker hub directly. 
I'll do a this with transition also to check if the Docker build there is also correct with this change.

greenscientist:
Correct build: https://hub.docker.com/repository/registry-1.docker.io/chairemobilite/trrouting/builds/6980879d-8fea-4794-ba2c-0172e2cb2e3a

greenscientist:
PR on the Transition side to also update to bookworm: https://github.com/chairemobilite/transition/pull/781

tahini (approved):
Ok then

PR #296, opened 2024-06-11
Check for index in path node list when parsing connection

greenscientist:
We had an issue when the data got corrupted and the number of entry in the path was not the same as in the schedule. This was corrupting the memory. We use the at() instead of [] when accessing the vector in the parsing loop in getSchedules() and we display a meaningful error message if we get the exceptiono

### DIFF
--- a/src/trips_and_connections_cache_fetcher.cpp
+++ b/src/trips_and_connections_cache_fetcher.cpp
@@ -102,9 +102,10 @@ namespace TrRouting
               for (unsigned long nodeTimeI = 0; nodeTimeI < nodeTimesCount - 1; nodeTimeI++)
               {
 
+                try {
                   connections.push_back(Connection(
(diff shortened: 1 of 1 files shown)


src/trips_and_connections_cache_fetcher.cpp
```diff
@@ -116,7 +117,10 @@ namespace TrRouting
                   ));
 
                   trip.connectionDepartureTimes[nodeTimeI] = departureTimesSeconds[nodeTimeI];
-
+                } catch (std::out_of_range const& exc) {
+                  spdlog::error("Index out of range will parsing connection for trip on line ({})", path.line.longname);
```
kaligrafy: while instead of will?

kaligrafy (changes requested):

tahini (approved):

kaligrafy (approved):

PR #297, opened 2024-06-12
Add a new error code for invalid scenario

### LINKED ISSUE #295: If the client send a wrong scenario, error message should state it so
Currently, if we send a wrong scenario ID, we get a MISSING_PARAM_SCENARIO, 
{'status': 'query_error', 'errorCode': 'MISSING_PARAM_SCENARIO'}

Instead of a message stating that the scenario was there but wrong.

greenscientist:
Currently, if we send a scenario id but that ID is not in the available cache data, we return the error MISSING_SCENARIO, which is missleading.

We add a new error code INVALID_SCENARIO, that will be returned if the scenario_id is not a valid uuid, or it's not present in our data source

Fixes: #295

### DIFF
--- a/connection_scan_algorithm/src/parameters/common_parameters.cpp
+++ b/connection_scan_algorithm/src/parameters/common_parameters.cpp
@@ -105,12 +105,20 @@ namespace TrRouting
       // scenario:
       else if (parameterWithValue.first == "scenario_id")
       {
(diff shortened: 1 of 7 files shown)


tahini (approved):

tahini:
When you merge, please add an issue in Transition to support this return value

greenscientist:
> When you merge, please add an issue in Transition to support this return value

Will do. Transition should never gets this error, so we'll need to make it quite apparent.

Issue #301, opened 2024-12-16
update for boost 1.87


kaligrafy:
1.87 fails because of an update in boost::asio:

../../include/client_http.hpp:20:44: error: no member named 'const_buffers_1' in namespace 'boost::asio'

greenscientist:
@kaligrafy , can you try boost 1.86 by any chance?

kaligrafy:
I can only install 1.76, 1.85 or 1.87 with brew

kaligrafy:
With boost@1.85, I get this error:
 boost::program_options::validators::check_first_occurrence(boost::any const&)
Undefined symbols for architecture arm64:

greenscientist:
It might be an issue only with arm64! (or simply mac/homebrew specifically) Tested on archlinux with both GCC and LLVM with boost 1.86 and it passes

greenscientist:
From boost 1.87 changelog: 
Removed deprecated classes const_buffers_1 and mutable_buffers_1.

So this is the issue.

greenscientist:
Problems seems to still be upstream in https://gitlab.com/eidheim/Simple-Web-Server

greenscientist:
There's other changes in https://gitlab.com/eidheim/Simple-Web-Server/-/merge_requests/280, but we don't seem to need it. 

I'll close this for now.

PR #303, opened 2024-12-20
Change simple-web to not use deprecated boost const_buffers_1

greenscientist:


greenscientist:
Closes #301

tahini (commented):
ok... and this Iterator is exactly equivalent and the same as previously? There is no other change required anywhere?

greenscientist:
Seems to work...  
I've done Claude's suggestion, it's the only changes to make it compile. 

I've tested in transition, accessbility map and route calculations works . 

Guessing that with this change the right iterator types is picked up. I've sent the changes upstream, wondering if it will be picked as is.

tahini (commented):
Which upstream? can you link to that PR to see what the upstream says? And fi there is an upstream, why do we have it here too?

Also, you change `client_http.hpp`, do you need to change `client_https.hpp` too? Or they are totally different files?

greenscientist:
Upstream PR: https://gitlab.com/eidheim/Simple-Web-Server/-/merge_requests/279

We do it here since for that project it's customary to copy the simple file into the project. We also have some local modifications I think. 

We don't need to change client_https.hpp, it uses the content of client_http.hpp

tahini (approved):
I you say so!

PR #306, opened 2025-02-13
Add workflow to build docker image on github

greenscientist:


greenscientist:
@tahini so far this is only tested for the test-docker-build file. 
Trying to see if we can test the other part without doing a commit first

greenscientist:
Tested by pushing a ghdockbuild branch and it worked well!

.github/workflows/build-docker-image.yml
```diff
@@ -0,0 +1,64 @@
...
+on:
+  push:
+    branches: 
+      - "v2c"
+  workflow_dispatch:
+
+env:
+  GITHUB_REGISTRY: ghcr.io
+  DOCKERHUB_REGISTRY: docker.io
+  DOCKERHUB_IMAGE: chairemobilite/${{ github.event.repository.name }}
```
tahini, line 12: Does it mean that it pushes both to dockerhub and github?

tahini (approved):
Sounds good?

.github/workflows/build-docker-image.yml
greenscientist: exactly!

Issue #271, opened 2023-11-10
accessibilityMap: v1 had an accessibleNodes parameter we could pass to restrict analyzed data


tahini:
This parameter was used in Transition's genetic algorithm to cache the accessible nodes from a place, in the accessibility indicator. It would speed up the tests.

Do we need/want to implement this again? Is this still an issue in simulations?

Keeping this as a not here to consider the question at some point.

greenscientist:
Some of that code was removed in https://github.com/chairemobilite/trRouting/commit/56e23abed9540dfe115390e63d8c570442dc8fe1
(if we need to reinstate it)

PR #308, opened 2025-03-24
Add support to cache accessible nodes via Memcached

greenscientist:
Related to issue #271
In the V1 API, Transition had the possibility to send a list of accessible nodes with a request. The intent for this was to speedup calculation for frequently requested points.

Instead of letting the client send us cache information, we are adding support for TrRouting to handle the caching directly using Memcached.

When enabled with --useMemcached command-line option, trRouting will compute a key with the point and caracteristic of the accessible node request (max walking time, speed, direction). If the data was already computed, it will be fetched from the memcached instance, otherwise it will do the calculation and store the result for later use.

It's implemented as a GeoFilter and receive a geofilter in its constructor, so it can be used with both the OSRMGeoFilter and the Euclidean one.

In our test, a request to OSRM was taking aroung 50-100ms, and with the caching we got the result in less than 1ms. Not a huge saving for a single calculation, but still a save. This will be mostly benefical with computations that often check the same points, like with Transition's simulations. Other type of batch calculation could gain from it if they have recuring access or egress points.

We currently cache the result for 1h in memcached, to not have stale results and conserve memory over time.

The compilation of this extension is optional. If you don't have libmemcached installed, it won't compile it. The CLI option is always available and will only display a warning if it was not compiled.

Tested with both OSRM and Euclidean filters, with and without Memcached (including the case when memcached was shutdown)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,8 +29,10 @@ brew install boost
 brew install capnp
 brew install spdlog
 brew install nlohmann-json
+brew install libmemcached
 ```
 
+libmemcached is optional
 ## Ubuntu 24.04 Install
 
 ```
@@ -41,6 +43,7 @@ You if you haven't installed other basic build dependencies, like autoconf, you 
 sudo apt install build-essential autoconf pkg-config
 ```
 
+libmemcached-dev is optional
 ## Compilation
 trRouting use autoconf/automake as its build system. A recap of the usual commands: 
 
--- a/README.md
+++ b/README.md
@@ -23,6 +23,16 @@ https://chairemobilite.github.io/trRouting/
 [5]: https://github.com/Project-OSRM/osrm-backend/blob/master/docs/profiles.md "OSRM profiles"
 [6]: https://github.com/Project-OSRM/osrm-backend/wiki/Running-OSRM "Running OSRM"
 
+### Memcached support
+trRouting have the ability to cache some of the accessible node calculation results. To do so, it can
+use an external memcached daemon.
+You tell trRouting to do this by passing the --useMemcached parameter. By default it will
+try to access memcached on the default port, 11211 on the localhost. You can pass an hostname/port
+to the useMemcached parameter like so: --useMemcached=localhost:11111
+
+The memcached support will only be compiled if the configure script can detect libmemcached on
+the system, hence why it's marked as optional in the instructions bellow
+
 ## Mac OS X Install with homebrew
 ```
 brew install boost
--- a/connection_scan_algorithm/src/program_options.cpp
+++ b/connection_scan_algorithm/src/program_options.cpp
@@ -34,6 +34,8 @@ namespace TrRouting {
       ("osrmCyclingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm cycling host");
     options.add_options()
       ("osrmDrivingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm driving host");
+    options.add_options()
+      ("useMemcached",                                     boost::program_options::value<std::string>()->implicit_value("localhost:11211"), "Enable memcached caching with optional server string");
 
   }
 
@@ -55,6 +57,8 @@ namespace TrRouting {
     osrmWalkingHost      = "localhost";
     osrmCyclingHost      = "localhost";
     osrmDrivingHost      = "localhost";
+    useMemcached         = false;
+    memcachedServers     = "localhost:11211";
 
     if(variablesMap.count("help")) {
       std::cout << options << std::endl;
@@ -133,6 +137,11 @@ namespace TrRouting {
     if(variablesMap.count("osrmDrivingHost") == 1)
     {
       osrmDrivingHost = variablesMap["osrmDrivingHost"].as<std::string>();
+    }
+    if(variablesMap.count("useMemcached") == 1)
+    {
+      useMemcached = true;
+      memcachedServers = variablesMap["useMemcached"].as<std::string>();
     }
 
   }
--- a/connection_scan_algorithm/src/program_options.cpp
+++ b/connection_scan_algorithm/src/program_options.cpp
@@ -35,7 +35,7 @@ namespace TrRouting {
     options.add_options()
       ("osrmDrivingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm driving host");
     options.add_options()
-      ("useMemcached",                                     boost::program_options::value<std::string>()->implicit_value("localhost:11211"), "Enable memcached caching with optional server string");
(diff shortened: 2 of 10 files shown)


greenscientist:
added commits for adding libmemcached to Dockerfile and README

kaligrafy:
Wow!

README.md
```diff
@@ -29,15 +29,18 @@ brew install boost
...
+libmemcached is optional
 ## Ubuntu 16.04 Install
 
 [Install Cap'nProto](https://capnproto.org/install.html)
 ```
-sudo apt-get install clang libboost-all-dev libexpat1-dev libjsoncpp-dev libspdlog-dev nlohmann-json3-dev
+sudo apt-get install clang libboost-all-dev libexpat1-dev libjsoncpp-dev libspdlog-dev nlohmann-json3-dev libmemcached-dev
 ```
 
+libmemcached-dev is optional
```
tahini, line 56: maybe explain why it is optional and why we would want it [or not]

greenscientist:
Need to add thread safety

greenscientist:
Thread safefy mecanism added for memcached

connection_scan_algorithm/src/program_options.cpp
```diff
@@ -34,6 +34,8 @@ namespace TrRouting {
       ("osrmCyclingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm cycling host");
     options.add_options()
       ("osrmDrivingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm driving host");
+    options.add_options()
+      ("useMemcached",                                     boost::program_options::value<std::string>()->implicit_value("localhost:11211"), "Enable memcached caching with optional server string");
```
tahini: A variable with this name, I would expect to be a boolean, `true` or `yes/no`, but it's actually the server:port that is expected? Maybe have another option for `--memCacheServers`?

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -137,6 +140,18 @@ int main(int argc, char** argv) {
     spdlog::info("Using OSRM for access/egress node time/distance");
   }
 
+  // Wrap the geoFilter with memcached if requested and available
+  if (programOptions.useMemcached) {
+    #ifdef HAVE_MEMCACHED
+      geoFilter =  new TrRouting::MemcachedGeoFilter(geoFilter, programOptions.memcachedServers);
```
tahini: So you do have another option for the server here... So, it is just not correctly defined in the `options` above?

src/memcachedgeofilter.cpp
```diff
@@ -0,0 +1,208 @@
+#include "memcachedgeofilter.hpp"
```
tahini, line 1: Our files don't have a copyright header? Shouldn't they?

tahini (changes requested):
I think the options are not properly documented. But looks great!

What would be the difficulty to add unit tests for this? At least for the serialization/de-serialization of the results by mocking the memcached library?

greenscientist:
> What would be the difficulty to add unit tests for this? At least for the serialization/de-serialization of the results by mocking the memcached library?

Quite difficult I think. We would need to write a fake library and link it to a test program that use the library

src/memcachedgeofilter.cpp
greenscientist: They probably should, but I imagine it's not really essential.

connection_scan_algorithm/src/program_options.cpp
greenscientist: I'm using the pattern that is common with configure parameters. You can say that you want to use it with some default parameters or use it with some specific parameter. 
The fact or using the parameter means that you want to use it. Saying --useMemcached=no does not make sense in our case, it's always no by default.

src/memcachedgeofilter.cpp
greenscientist: I'm surprised no files seem to have it!

connection_scan_algorithm/src/program_options.cpp
greenscientist: Here's the --help output now: 
  --useMemcached [=arg(=localhost:11211)]
                                        Enable memcached caching. You can 
                                        specify a non-default memcached server 
                                        to use by adding an optional 
                                        hostname:port string

src/memcachedgeofilter.cpp
tahini: > I'm surprised no files seem to have it!

I was surprised too! d'où mon commentaire

README.md
```diff
@@ -23,21 +23,34 @@ https://chairemobilite.github.io/trRouting/
 [5]: https://github.com/Project-OSRM/osrm-backend/blob/master/docs/profiles.md "OSRM profiles"
 [6]: https://github.com/Project-OSRM/osrm-backend/wiki/Running-OSRM "Running OSRM"
 
+### Memcached support
+trRouting have the ability to cache some of the OSRM results it use. To do so, it can
```
tahini: OSRM results => accessible node calculation results, puisque ça fonctionnerait aussi avec les euclidian distance je crois.

connection_scan_algorithm/src/program_options.cpp
```diff
@@ -34,6 +34,8 @@ namespace TrRouting {
       ("osrmCyclingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm cycling host");
     options.add_options()
       ("osrmDrivingHost",                                   boost::program_options::value<std::string>()->default_value("localhost"), "osrm driving host");
+    options.add_options()
+      ("useMemcached",                                     boost::program_options::value<std::string>()->implicit_value("localhost:11211"), "Enable memcached caching. You can specify a non-default memcached server to use by adding an optional hostname:port string");
```
tahini, line 38: Donc juste `--useMemcached` sans le `=server:port` fonctionnerait aussi et utiliserait la valeur par défaut?

connection_scan_algorithm/src/program_options.cpp
```diff
@@ -134,6 +138,11 @@ namespace TrRouting {
     {
       osrmDrivingHost = variablesMap["osrmDrivingHost"].as<std::string>();
     }
+    if(variablesMap.count("useMemcached") == 1)
+    {
+      useMemcached = true;
+      memcachedServers = variablesMap["useMemcached"].as<std::string>();
```
tahini, line 144: Que fait cette ligne si on a mis juste `--useMemcached` sans valeur?

connection_scan_algorithm/src/program_options.cpp
greenscientist: Oui exact.

connection_scan_algorithm/src/program_options.cpp
greenscientist: Ca prend la valeur qui a été définie dans le implicit_value("localhost:11211")

tahini (approved):

PR #322, opened 2025-09-22
route: calculate route with arrival time symetrically with departure

### LINKED ISSUE #298: Strange itineraries when routing based on arrival time
Routing using the _Arrival time_ parameter seems to introduce distorsions in the trip itinerary chosen by trRouting. For instance:

Here is a random trip using a random starting time (7:11). Notice how the proposed itinerary below includes a 3-minute walk from the last stop for a total trip time of **41 minutes** and an arrival at 7:52.
image

If we switch to using the arrival time and set it to 7:53 (to account for any rounding), we get the itinerary below. Notice how the last stop changes to the next one on the route, but we still get a 3-minute egress walk and a total trip time of **41 minutes**, arriving at 7:52.
image

Now, if we set the arrival time to 8:00, we get the strange behavior illustrated below. The starting time is the same as the 2 other scenarios, but the arrival time is 4 minutes later at 7:56, for at total trip time of **46 minutes**. Notice how the egress stop is farther down the route and the itinerary then backtracks on foot via a 7-minute walk. 
image

It seems like trRouting tries to arrive as close as possible to the _Arrival time_ parameter, even if that introduces extra vehicle time and walking time.

tahini:
fixes #298

Fix a time instead of distance in the `forwardJourneyStep` method.

Use a similar approach with arrival time calculations as the departure time one: first get the best access time by a reverse calculation, then find the best arrival time for this departure time, thus optimizing the arrival of the route. Then retrieve the journey steps required for the route.

## Summary by CodeRabbit

- **New Features**
  - Added a dedicated forward calculation path for single-route queries.
  - Introduced local journey optimization profiling to inform route selection.

- **Bug Fixes**
  - Improved boarding and transfer selection via transfer-time-based reassignment.
  - Corrected time metric usage in journey steps for more accurate timings.
  - Clarified behavior when no route is found.

- **Tests**
  - Added an arrival-time scenario with two alternatives to validate preferred transfer routing.

### DIFF
--- a/connection_scan_algorithm/src/forward_journey.cpp
+++ b/connection_scan_algorithm/src/forward_journey.cpp
@@ -44,7 +44,6 @@ namespace TrRouting
       int waitingTime              {-1}; int accessWaitingTime      {-1};
 
       // recreate journey:
-
       JourneyStep resultingNodeJourneyStep = forwardEgressJourneysSteps.at(resultingNode.uid);
 
       std::optional<std::reference_wrapper<const Node>> bestAccessNode;
@@ -58,7 +57,7 @@ namespace TrRouting
       journey.push_back(JourneyStep(std::nullopt,
                                     std::nullopt,
                                     std::nullopt,
-                                    nodesEgress.at(resultingNode.uid).distance,
+                                    nodesEgress.at(resultingNode.uid).time,
                                     false,
                                     nodesEgress.at(resultingNode.uid).distance));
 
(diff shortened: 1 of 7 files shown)


tahini:
This does the same thing as #316, but instead of doing 3 passes just to re-use the same logic as forward calculation, this does a symetrical to departure time 2-pass calculation.

tahini:
Prêt pour review, le 1er commit, c'est #324 qui peut être reviewé séparément

tahini:
When I have a few minutes, I'll try to see if this fixes any of https://github.com/chairemobilite/transition/issues?q=state%3Aopen%20label%3A%22transit%20alternatives%22

connection_scan_algorithm/src/forward_journey.cpp
```diff
@@ -58,7 +57,7 @@ namespace TrRouting
       journey.push_back(JourneyStep(std::nullopt,
                                     std::nullopt,
                                     std::nullopt,
-                                    nodesEgress.at(resultingNode.uid).distance,
+                                    nodesEgress.at(resultingNode.uid).time,
```
greenscientist, line 60: That seems like a pretty important fix!

greenscientist (approved):
Ok, seems fine. 

I really feel like we should be able to merge the forward and reverse code somehow. Or at least have more function, but that's work for later.

PR #323, opened 2025-09-24
csa: update a few debug statements in alternative calculation

tahini:
* Move the failed combinations at the end only instead of at every step
* Add the travel time seconds where it should be
* Display the number of alternatives calculated instead of the max

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured all failed combinations are reported after processing completes, improving visibility of outcomes.
* **Refactor**
  * Moved reporting of failed combinations to post-processing to reduce noisy iteration logs and clarify flow.
* **Chores**
  * Clarified travel time log wording to include explicit units.
  * Expanded final progress message to include a "done" status and maximum possible alternatives for clearer status context.

### DIFF
--- a/connection_scan_algorithm/src/alternatives_routing.cpp
+++ b/connection_scan_algorithm/src/alternatives_routing.cpp
@@ -269,26 +269,25 @@ namespace TrRouting
         }
 
         alternativesCalculatedCount++;
-
-        if (failedCombinations.size() > 0)
-        {
-          for (auto failedCombination : failedCombinations)
-          {
-            spdlog::debug("failed combinations: {}", LinesToString(failedCombination));
-          }
-        }
       }
     }
 
     int i {0};
     for (auto flines : alreadyFoundLines)
     {          
-      spdlog::debug("{}. {} tt: ", i, LinesToString(flines.first),
+      spdlog::debug("{}. {} travel time minutes: {}", i, LinesToString(flines.first),
                     (foundLinesTravelTimeSeconds[flines.first] / 60));
       i++;          
     }
+
+    // Print failed combinations
+    for (auto failedCombination : failedCombinations)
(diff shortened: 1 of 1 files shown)


connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -269,26 +269,27 @@ namespace TrRouting
...
     for (auto flines : alreadyFoundLines)
     {          
-      spdlog::debug("{}. {} tt: ", i, LinesToString(flines.first),
+      spdlog::debug("{}. {} travel time seconds: {}", i, LinesToString(flines.first),
                     (foundLinesTravelTimeSeconds[flines.first] / 60));
       i++;          
     }
+
+    // Print failed combinations
+    if (failedCombinations.size() > 0)
```
greenscientist: In theory, we don't even need this if, for the would not loop.

connection_scan_algorithm/src/alternatives_routing.cpp
```diff
@@ -269,26 +269,27 @@ namespace TrRouting
...
+    {
+      for (auto failedCombination : failedCombinations)
+      {
+        spdlog::debug("failed combinations: {}", LinesToString(failedCombination));
+      }
+    }
     
-    spdlog::debug("last alternative found at: {} on a total of {} calculations", lastFoundedAtNum,  maxAlternatives);
+    spdlog::debug("last alternative found at: {} on a total of {} calculations done", lastFoundedAtNum,  alternativesCalculatedCount);
```
greenscientist: Maybe we'd want to keep maxAlternative? (If we make it a parameter someday)

greenscientist (approved):
You don't need to change things, only if you feel like it

PR #324, opened 2025-09-24
csa: throw no routing found in calculator if no best time

tahini:
Instead of letting the `reverseJourneyStep` function do it, as we already know this is what is going to happen, just throw the exception right away.

## Summary by CodeRabbit

* **Bug Fixes**
  * When no route is found the app now returns a clear error instead of triggering an internal assertion, preventing crashes and improving reliability in edge cases.
  * Removed unused fallback processing and noisy timing/debug logs in this scenario to reduce log clutter and avoid misleading diagnostics.
  * Ensures consistent propagation of “no route found” outcomes across the app for better user feedback and overall stability.

### DIFF
--- a/connection_scan_algorithm/src/calculator.cpp
+++ b/connection_scan_algorithm/src/calculator.cpp
@@ -32,45 +32,32 @@ namespace TrRouting
 
     if (departureTimeSeconds > -1 && parameters.isForwardCalculation())
     {
-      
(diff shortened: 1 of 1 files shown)


tahini:
@greenscientist j'ai ajouté le log statement demandé par lapinot

greenscientist (approved):
Yeah, this looks good.

PR #332, opened 2026-01-09
Add memcached cache persistence with node validation

kaligrafy:
Implement file-based persistence for memcached footpath cache data:

- Add single-file persistence (default: footpaths.cache) that saves all cached access/egress footpath data to disk
- Implement signal handling (SIGINT/SIGTERM) to automatically save cache on graceful shutdown
- Add node validation: cache stores hash of node UUIDs and locations, invalidates automatically if transit nodes change
- Set memcached TTL to 0 (never expires) for consistent caching behavior

New API endpoints:
- /saveCache - manually save cache to disk
- /loadCache - reload cache from disk
- /resetCache - flush memcached, clear local cache, delete cache file
- /cacheStatus - get cache statistics and persistence state

Cache file format:
- Magic header "TRFC" (TrRouting Footpath Cache) with version number
- Stores node count and hash for validation on reload
- Binary serialization using Boost for efficiency

Update README with memcached setup instructions (including macOS/brew), cache persistence documentation, and API endpoint reference.

Add comprehensive tests for persistence, node validation, cache file format, and API route patterns.

## Release Notes

* **New Features**
  * Memcached now supports disk-based cache persistence to preserve routing data across restarts.
  * Added new cache management endpoints for manual control: save, load, reset, and status queries.
  * Implemented graceful shutdown that automatically persists cache when enabled.

* **Documentation**
  * Expanded memcached setup guide with platform-specific installation instructions, configuration examples, cache persistence details, and cache invalidation scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


greenscientist (changes requested):
I'm declining that right away. It's a bad idea. Doing cache persistance just defeat the purpose of a cache, it then became a database with all the complexity you have implemented here.

kaligrafy:
> I'm declining that right away. It's a bad idea. Doing cache persistance just defeat the purpose of a cache, it then became a database with all the complexity you have implemented here. 

Ok but what do you propose? Right now the simulations using trrouting are using 10 X more time to refetch the accessible nodes (access/egress)  every time for each demand trip than to calculate the transit route itself, making it non usable. And you mentioned we should not save the OD trips to the db either for now.

kaligrafy:
See https://github.com/chairemobilite/transition/issues/1685

PR #338, opened 2026-03-04
Reduce memory churn by not calling shrink_to_fit

greenscientist:
On an high thread count, we are getting contention on memory by constantly allocating and deallocating memory. We can save some of that by not calling shrink_to_fit. We might use a bit more memory, but gain on performance

## Summary by CodeRabbit

* **Refactor**
  * Adjusted memory management strategy to reduce explicit memory capacity releases during data resets and cache operations.

### DIFF
--- a/connection_scan_algorithm/src/resets.cpp
+++ b/connection_scan_algorithm/src/resets.cpp
@@ -29,9 +29,7 @@ namespace TrRouting
     if (resetAccessPaths)
     {
       accessFootpaths.clear();
-      accessFootpaths.shrink_to_fit();
       egressFootpaths.clear();
-      egressFootpaths.shrink_to_fit();
     }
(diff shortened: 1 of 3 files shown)


greenscientist:
We might need to benchmark this a bit more

tahini:
As-tu roulé les benchmarks sur cette PR?

greenscientist:
> As-tu roulé les benchmarks sur cette PR?

Il faut que j'en refasse, maintenant qu'on a identifié le bottleneck to dispatch http.

kaligrafy (approved):

tahini (changes requested):
Putting a request changes to not merge accidentally until we know what the benchmarks say and confirm it is needed

greenscientist:
relaced by #363

PR #350, opened 2026-07-15
Update googletest submodule to v1.17.0

greenscientist:
## Summary by CodeRabbit

* **Chores**
  * Updated the bundled GoogleTest revision to a newer snapshot.


greenscientist:
@coderabbitai configuration

greenscientist:
@coderabbitai review

tahini (commented):
Since we don't do that often, you could add the origin version in the commit message, just so we can appreciate if we update the major, minor, or other.

greenscientist:
> Since we don't do that often, you could add the origin version in the commit message, just so we can appreciate if we update the major, minor, or other.

Done. The update was from 1.12.1

tahini (approved):

PR #362, opened 2026-07-27
Disable idle connection kickoff in drogon

greenscientist:
By default, drogon would kill a connection that is idle for 60 seconds. It's quite common for calculation to take more time than that, so we don't want this behavior.

We could consider in the future to set it to a really high value.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of long-running transit routing requests by keeping idle HTTP connections open long enough for calculations to finish, reducing premature disconnects during inactivity.

### DIFF
--- a/connection_scan_algorithm/src/transit_routing_http_server.cpp
+++ b/connection_scan_algorithm/src/transit_routing_http_server.cpp
@@ -470,6 +470,9 @@ int main(int argc, char** argv) {
   drogon::app()
     .addListener("0.0.0.0", programOptions.port)
     .setThreadNum(4) // IO event loops only; calculations run on the compute pool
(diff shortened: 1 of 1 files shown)


kaligrafy (changes requested):
Je sais pas

kaligrafy:
I would put a very big value so stuck connection do timeout eventually

greenscientist:
> I would put a very big value so stuck connection do timeout eventually

So what is the really big max of a calculation you can imagine for trRouting to do ?

kaligrafy:
I don't know. There is no way to check if trrouting is indeed calculating or just stuck?

greenscientist:
> I don't know. There is no way to check if trrouting is indeed calculating or just stuck?

CPU usage and timing clearly shows that trrouting was calculating

kaligrafy (approved):

PR #363, opened 2026-07-28
Use a permanent thread local Calculation object

greenscientist:
One of the main performance issue we have is the memory allocation churn of creating and delete the Calculation object for each request. Instead of always reallocating for each object, we can use a local object for each thread. Since we only have one request at a time being handled on a thread, we don't have concurrency issues.

We have to call reset at the start of each request to clear out the data.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved transit routing request processing by reusing per-thread calculation resources across requests instead of creating new instances each time.
  * Added request-level timing so performance metrics capture each request’s calculation duration more accurately.
* **Bug Fixes**
  * Improved calculation state cleanup during resets by clearing egress footpaths and access/egress node state earlier, reducing the risk of stale routing data affecting subsequent computations.

### DIFF
--- a/connection_scan_algorithm/src/transit_routing_http_server.cpp
+++ b/connection_scan_algorithm/src/transit_routing_http_server.cpp
@@ -150,6 +150,23 @@ static void runCalculation(trantor::ConcurrentTaskQueue &computePool,
     });
 }
 
+// Return the Calculator owned by the calling compute thread, creating it on
+// first use. Constructing a Calculator per request allocated (and immediately
+// freed) network-sized containers, which dominated system time under load.
+// The compute pool threads live for the whole process, and a task always runs
+// to completion on a single thread, so one instance per thread is safe and
+// lets reset() reuse warm allocations instead of faulting in fresh pages.
+// Note: this must not be called from the Drogon IO loop threads, only from
+// tasks queued on the compute pool.
+static Calculator & getThreadCalculator(const TransitData &transitData, GeoFilter &geoFilter)
+{
+  thread_local Calculator calculator(transitData, geoFilter);
+  // Start the timer at the beginning of each request.
+  // TODO Maybe we should move the timer out of the calculator
+  calculator.startRequestTimer();
+  return calculator;
+}
+
 // Register a GET handler for both /path and /path/ to keep the behavior of
 // the previous "[/]?" route regexes
 static void registerGet(const std::string &path, Handler handler)
@@ -340,7 +357,7 @@ int main(int argc, char** argv) {
 
       spdlog::info("-- calculating route request -- {}", currentRequestId);
 
-      Calculator calculator(transitData, *geoFilter);
+      Calculator &calculator = getThreadCalculator(transitData, *geoFilter);
       RouteParameters queryParams = RouteParameters::createRouteODParameter(parametersWithValues, transitData.getScenarios());
 
       try {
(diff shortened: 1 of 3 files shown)


connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -150,6 +150,23 @@ static void runCalculation(trantor::ConcurrentTaskQueue &computePool,
...
+// tasks queued on the compute pool.
+static Calculator & getThreadCalculator(const TransitData &transitData, GeoFilter &geoFilter)
+{
+  thread_local Calculator calculator(transitData, geoFilter);
+  // Start the timer at the beginning of each request.
+  // TODO Maybe we should be it out of the calculator
+  calculator.startRequestTimer();
+  return calculator;
+}
+
```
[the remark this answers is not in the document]
greenscientist: For the moment, we either start a new trRouting on batch calculation or restart most instances after a few days, so this will not be an issue for the short/medium term

kaligrafy (approved):

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -150,6 +150,23 @@ static void runCalculation(trantor::ConcurrentTaskQueue &computePool,
...
+// The compute pool threads live for the whole process, and a task always runs
+// to completion on a single thread, so one instance per thread is safe and
+// lets reset() reuse warm allocations instead of faulting in fresh pages.
+// Note: this must not be called from the Drogon IO loop threads, only from
+// tasks queued on the compute pool.
+static Calculator & getThreadCalculator(const TransitData &transitData, GeoFilter &geoFilter)
+{
+  thread_local Calculator calculator(transitData, geoFilter);
+  // Start the timer at the beginning of each request.
+  // TODO Maybe we should be it out of the calculator
```
tahini: This sentence does not make sense: "be it out"

connection_scan_algorithm/src/transit_routing_http_server.cpp
```diff
@@ -150,6 +150,23 @@ static void runCalculation(trantor::ConcurrentTaskQueue &computePool,
...
+// first use. Constructing a Calculator per request allocated (and immediately
+// freed) network-sized containers, which dominated system time under load.
+// The compute pool threads live for the whole process, and a task always runs
+// to completion on a single thread, so one instance per thread is safe and
+// lets reset() reuse warm allocations instead of faulting in fresh pages.
+// Note: this must not be called from the Drogon IO loop threads, only from
+// tasks queued on the compute pool.
+static Calculator & getThreadCalculator(const TransitData &transitData, GeoFilter &geoFilter)
+{
+  thread_local Calculator calculator(transitData, geoFilter);
```
tahini, line 163: C'est le fait que la fonction est statique qui fait que la variable définie comme thread_local à l'intérieur existe durant toute la durée du thread? Je ne connais pas ce keyword... La variable n'est pas définie "globalement/par thread", elle est quand même dans cette fonction après tout.

greenscientist: https://en.cppreference.com/cpp/language/storage_duration

Thread storage duration
All variables declared with thread_local have thread storage duration.

The storage for these entities lasts for the duration of the thread in which they are created. There is a distinct object or reference per thread, and use of the declared name refers to the entity associated with the current thread.

Elle est dans la fonction, mais l'objet existe pour toute la durée du thread. C'est un peu comme statique oui, mais pour chaque thread.

connection_scan_algorithm/src/transit_routing_http_server.cpp
greenscientist: Je trouve ca étrange que le lapin aille pas vu ça!

connection_scan_algorithm/src/transit_routing_http_server.cpp
tahini: Si tu le dis...

tahini (approved):

PR #373, opened 2026-08-05
Refactor the line alternative filter management

### LINKED ISSUE #371: Move alternative line filter to its own data structure instead of changing query parameter
The parameter object has options for filtering on almost every object types, to mimic the filter in scenarios. 

The alternative calculation plays with this to exclude lines available for the calculation. Instead of changing the query params, we should just use a different parameters dedicated to alternatives.

greenscientist:
We were piggybacking on the original query parameter to add an extra line filter calculate call. We had to expose publicly the exclude line parameter of the object to make this work.

Instead of relying on a quirk of the original paramater, we pass a new AlternativeFilter object to the calculateSingle call. This is passed to the reset call, in place of the usused resetFilter which was always true.

We made a generic interface AlternativeFilter that can be implemented in various way. We currently have an AlternativeLineFilter implemented.

This change make the management of the alternative filtering more obvious and will make it easier to implement variant of the alternative generation algorithm.

## Summary by CodeRabbit

* **New Features**
  * Added support for filtering alternative routes by excluding selected transit lines.
  * Alternative route calculations now apply line exclusions through a dedicated filtering mechanism.

* **Improvements**
  * Improved route recalculation by consistently resetting disabled trips before applying filters.
  * Restricted direct access to line-exclusion settings for safer configuration.
  * Added reliable handling for empty, overlapping, cumulative, and repeated line exclusions.

### DIFF
--- a/connection_scan_algorithm/include/alternative_filter.hpp
+++ b/connection_scan_algorithm/include/alternative_filter.hpp
@@ -1,0 +1,47 @@
+#pragma once
+
+#include <unordered_map>
+#include <vector>
+
+#include "line.hpp"
+#include "connection_set.hpp"
+#include "trip.hpp"
+
+namespace TrRouting
+{
+
+  /**
+     Generic interface to handle configuring and executing a filter to disable some trips to generate alternatives
+     route results
+   */  
+  class AlternativeFilter {
+  public:
+    AlternativeFilter() {}
+    virtual ~AlternativeFilter() = default;
+    /**
+       Apply a filter to a connectionSet and set flags in the trips disabled overlay. The clearing
+       of the trips overlay should be done outside of the filter, so we can apply multiple filter to the same set
+
+       @param tripsDisabled Container to be filled with flags of which trips are no longer available
+       @param connectionSet Current list of trips matching the query scenarios
+
+       @return void (Could be changed to return a bool to represent if something was filtered or not)
+     */
+    virtual void runFilter(std::unordered_map<Trip::uid_t, bool> &tripsDisabled, const ConnectionSet &connectionSet) = 0;
+  };
+
+  /** Alternative Line filter
+      Take a list of line and will disable all trips using that line.
+   */
+  class AlternativeLineFilter : public AlternativeFilter {
+  public:
+    AlternativeLineFilter(const std::vector<std::reference_wrapper<const Line>> &_excludedLines);
+    virtual void runFilter(std::unordered_map<Trip::uid_t, bool> &tripsDisabled, const ConnectionSet &connectionSet) override;
+  protected:
+    /** Simplified copy of the line vector (with only the UIDs) */
+    std::vector<Line::uid_t> excludedLinesUids;
+  };
+    
+}
+
+
--- a/connection_scan_algorithm/src/alternative_filter.cpp
+++ b/connection_scan_algorithm/src/alternative_filter.cpp
@@ -1,0 +1,37 @@
+#include "alternative_filter.hpp"
+
+#include <algorithm>
+
+#include "line.hpp"
+#include "connection_set.hpp"
+#include "trip.hpp"
+
+namespace TrRouting {
+
+  AlternativeLineFilter::AlternativeLineFilter(const std::vector<std::reference_wrapper<const Line>> &_excludedLines)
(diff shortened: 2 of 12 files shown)


connection_scan_algorithm/include/alternative_filter.hpp
```diff
@@ -0,0 +1,30 @@
...
+  public:
+    AlternativeFilter() {}
+    virtual ~AlternativeFilter() = default;
+    virtual bool runFilter(std::unordered_map<Trip::uid_t, bool> &tripsDisabled, const ConnectionSet &connectionSet) = 0;
+  };
+    
+  class AlternativeLineFilter : public AlternativeFilter {
+  public:
+    AlternativeLineFilter(const std::vector<std::reference_wrapper<const Line>> &_excludedLines);
+    virtual bool runFilter(std::unordered_map<Trip::uid_t, bool> &tripsDisabled, const ConnectionSet &connectionSet) override;
```
tahini: svp documenter ces paramètres. tripsDisabled, c'est disabled avant ou après la fonction? C'est quoi les attentes par rapport à ces paramèters (pré-conditions, modifications)

connection_scan_algorithm/src/alternative_filter.cpp
```diff
@@ -0,0 +1,43 @@
...
+    for (const Line & line : _excludedLines) {
+      excludedLinesUids.push_back(line.uid);
+    }
+    
+  }
+  
+  bool AlternativeLineFilter::runFilter(std::unordered_map<Trip::uid_t, bool> &tripsDisabled, const ConnectionSet &connectionSet) {
+
+    // Make sure the container is empty
+    tripsDisabled.clear();
```
tahini: tripDisabled est donc un output? Pourquoi ne pas le retourner comme tel au lieu d'un booléen toujours true et modifier une liste qui doit avoir été créée par l'appelant?

include/line.hpp
```diff
@@ -2,6 +2,7 @@
 #define TR_LINE
 
 #include <string>
+#include <memory>
```
tahini, line 5: Why this new include?

include/trip.hpp
```diff
@@ -3,6 +3,7 @@
 
 #include <boost/uuid/uuid.hpp>
 #include <vector>
+#include <optional>
```
tahini, line 6: Is this new import needed?

include/line.hpp
greenscientist: yup, its a required one for the file that was not explicitely there

include/trip.hpp
greenscientist: yup, its a required one for the file that was not explicitely there

connection_scan_algorithm/src/alternative_filter.cpp
greenscientist: car la structure n'est pas créé par la fonction, on ne veut pas de nouvelle allocation mémoire pour éviter un churn inutile avec l'OS.

connection_scan_algorithm/include/alternative_filter.hpp
greenscientist: Oui, oui, la PR n'est pas encore terminé.

tests/connection_scan_algorithm/alternative_filter_test.cpp
```diff
@@ -0,0 +1,280 @@
...
+  std::unordered_map<TrRouting::Trip::uid_t, bool> tripsDisabled;
+  filter.runFilter(tripsDisabled, *connectionSet);
+  ASSERT_EQ(2u, tripsDisabled.size());
+
+  filter.runFilter(tripsDisabled, *connectionSet);
+
+  EXPECT_EQ(2u, tripsDisabled.size());
+  EXPECT_TRUE(isDisabled(tripsDisabled, getTrip(TestDataFetcher::trip1EWUuid)));
+  EXPECT_TRUE(isDisabled(tripsDisabled, getTrip(TestDataFetcher::trip2EWUuid)));
+}
```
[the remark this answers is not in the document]
greenscientist: That's not something that we have a good setup here, but the overall correctness should be covered by the alternative calculation test

tahini (approved):

PR #374, opened 2026-08-06
Big set of small clean up fixes

greenscientist:
See each commit for a detailled description

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trip comparison consistency using stable local identifiers.
  * Added reliable ordering and inequality comparisons for trips.

* **Performance**
  * Streamlined route, connection, and transit data processing while preserving routing behavior.

* **Maintenance**
  * Improved parameter handling across routing and accessibility calculations.
  * Simplified internal processing and removed obsolete diagnostics, warnings, duplicate references, and unused code.

### DIFF
--- a/include/trip.hpp
+++ b/include/trip.hpp
@@ -46,7 +46,10 @@ namespace TrRouting
 
     uid_t uid; //Local, temporary unique id, used to speed up lookups
 
-    inline bool operator==(const Trip& other ) const { return uuid == other.uuid; }
+    // Equal operator. We only compare the uid, since they are unique.
+    inline bool operator==(const Trip& other ) const { return uid == other.uid; }
+    inline bool operator<(const Trip& other ) const { return uid < other.uid; }
+    inline bool operator!=(const Trip& other ) const { return uid != other.uid; }
 
     static uid_t getMaxUid() { return global_uid; }
   private:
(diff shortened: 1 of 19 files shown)


greenscientist:
I found half of them  and then asked Claude to find me a few more.

include/trip.hpp
```diff
@@ -46,7 +46,10 @@ namespace TrRouting
 
     uid_t uid; //Local, temporary unique id, used to speed up lookups
 
-    inline bool operator==(const Trip& other ) const { return uuid == other.uuid; }
+    // Equal operator. We only compare the uid, since they should be unique.
```
tahini: They are or they should be?

include/trip.hpp
```diff
@@ -46,7 +46,10 @@ namespace TrRouting
...
+    inline bool operator==(const Trip& other ) const { return uid == other.uid; }
```
tahini, line 50: Also commit message says "to match Node and Line structure", but the commit detail is for Trip structure...

tahini (approved):
Looks good. Just a slight mismatch of a commit message and the code that it represents.

include/trip.hpp
greenscientist: Yes, Node and and Line already have those operators, so we do the same for Trip. Any suggestion for better wording?

include/trip.hpp
greenscientist: They are unless we made a mistake? :)

greenscientist:
adjusted comment and commit message

PR #381, opened 2026-08-14
alternatives: consolidate loop condition and max alternative check

greenscientist:
Small optimisation, the loop was iterating through all the combination, even if we had reached the max alternative count. The if was evaluated and skipped the rest of the calculation, but we were still looping.

## Summary by CodeRabbit

* **Performance**
  * Improved alternative route processing by stopping earlier when configured limits are reached.
  * Reduced redundant calculations during route alternative evaluation.
* **Behavior**
  * Route calculation results, combinations, and failure handling remain unchanged.


greenscientist:
I was getting up to 1% improvement in some of the benchmarks

tahini (approved):

